PR #23, opened 2020-03-28
legends in MakieLayout

piever:
A bit of a hack, as of now, in case of multiple legends the "legendelement" for each entry is the superimposition of all the traces that respect it. I guess it should really be the style that one gets when ignoring the other "primary keys", but that may be a bit problematic as that is something that is never plotted. Maybe one can plot those on a fake scene to get the legend elements?

Separate issue: there is no `legendelement` for `ShadedLine`, that is used in StatsMakie for linear regressions with error bars. Not sure where the implementation should live. Interestingly, there is actually no need for composite traces like `ShadedLine` if one uses the algebra of graphics (it's simply `Band + Lines`), so maybe the `Analysis` from StatsMakie could actually be ported here.

DEMO:

```julia
using AlgebraOfGraphics: dims, primary, data, spec, table, layoutplot
using AbstractPlotting, GLMakie, MakieLayout
using RDatasets: dataset
iris = dataset("datasets", "iris")
iris.New = rand(Bool, 150)
d = data([:SepalLength, :SepalWidth], [:PetalLength :PetalLength])
grp = primary(color = :Species, marker = :New) * primary(layout_x = dims(1), layout_y = dims(2))
s = spec(Scatter, markersize = 5px)
table(iris) * d * grp * s |> layoutplot
```
scatter

Also, to be discussed: how to make the legend placement customizable? I imagine by default, one could return the `scene` object, the `layout` object and the list of legends, and the user decides what to make with them?

cc: @jkrumbiegel (as I don't really know what I'm doing with MakieLayout).


jkrumbiegel:
Here is an example of a legend like it would result from a facet plot. I think it's best to make one subgridlayout just for the facets, and a container to put the legend in (here the container is just the topmost grid layout, but in your function I would create a separate one). This way the whole facet plot is self contained and the legend doesn't mess with the facets in case those are changed (by observables for example).

```julia
scene, layout = layoutscene()

facetlayout = layout[1, 1] = GridLayout()

axs = facetlayout[1:4, 1:3] = [LAxis(scene) for i in 1:4, j in 1:3]
linkaxes!(axs...)

markers = [:cross, :circle, :x]
colors = [:red, :blue, :green]

for ax in axs
    scatter!(ax, randn(30), color = rand([:red, :blue, :green], 30), marker = rand(markers, 30), markersize = 20px)
end

markerelements = [MarkerElement(color = :black, strokecolor = :transparent, marker = m) for m in markers]
colorelements = [PolyElement(color = c, strokecolor = :transparent) for c in colors]

legend = layout[1, 2] = LLegend(scene,
    [markerelements, colorelements],
    [["Group A", "Group B", "Group C"], ["Red", "Blue", "Green"]],
    ["Markers", "Colors"])

hidexdecorations!.(axs[1:end-1, :])
hideydecorations!.(axs[:, 2:end])

scene
```

grafik

> Separate issue: there is no legendelement for ShadedLine, that is used in StatsMakie for linear regressions with error bars. Not sure where the implementation should live.

That's a good question, I think recipes should additionally define what they want their legend symbols to look like. With MarkerElements, PolyElements and LineElements you can almost make any composite LegendElement, because you are allowed to stack multiple elements and also change their marker positions, poly points, etc.

piever:
I see, I can definitely implement your approach, I'll try to clean things up in the next few days.

> With MarkerElements, PolyElements and LineElements you can almost make any composite LegendElement, because you are allowed to stack multiple elements and also change their marker positions, poly points, etc.

Yes, definitely, this should really just be a line with a poly around it in the case of `ShadedLine`. My main concern is the dependency structure, because StatsMakie defines `ShadedLine` but does not depend on MakieLayout. This probably will need to wait for the Makie / MakieLayout refactor.

jkrumbiegel:
Yeah as this whole legend stuff should be in AbstractPlotting, I think you can already think about how to use it within StatsMakie. I've already factored out GridLayoutBase and the rest of MakieLayout.jl is really more like an AbstractPlotting extension with layoutable recipes.

piever:
I get the result I want now:

scatter

other than the issue with the y label being so far off. I imagine it is to allow for longer tick labels, should they change with the zoom? Is there an option to make the spacing tighter?

jkrumbiegel:
Yes there is tight_ticklabel_spacing!(ax), the ticklabels can't be directly tied to the layout because of interactivity. Maybe I can find a compromise but for now you have to do this manually

PR #34, opened 2020-04-05
choropleth support

piever:


piever:
Example:

```julia
using GeoJSON
states = download("https://raw.githubusercontent.com/PublicaMundi/MappingAPI/master/data/geojson/us-states.json") 
states_geo = GeoJSON.read(read(states, String))
data(states_geo) * style(color = :density) * spec(Poly, colorrange = (0, 1000)) |> draw
```

statesdensity

```julia
data(states_geo) * group(color = :density) * spec(Poly, colorrange = (0, 1000)) |> draw
```
statesname

PR #35, opened 2020-04-05
initial support continuous scales

piever:


piever:
TODO: make sure to allow users to pass custom scales, and add a "palette of continuous scales" as well.

PR #39, opened 2020-04-28
use CairoMakie release

piever:


asinghvi17:
I should mention that scatter markers will be wrong here, because the marker offset is not respected.  I'll let you know when that's fixed (I have the fix, just need to merge it).

PR #43, opened 2020-05-03
leave grid when hiding decorations

jkrumbiegel:
I changed the way these functions work by default at some point, but in the way used here it's better to keep the grid


piever:
Looks good, thanks!

PR #46, opened 2020-07-07
improve faceting

### LINKED ISSUE #42: Facet vs layout
Probably faceting and layouting could be two different options in this package. Faceting is for grouping data, layouting for plotting different variables (as in your example in the docs)

First, layouting.

```
N = 500
x = rand(N)
gx = rand(["x grp 1","x grp 2", "x grp 3"], N)
gy = rand(["y grp 1","y grp 2"], N)

df = DataFrame(g = g, x = x, y = (g .== "very") .+ x .+ randn(N), gx=gx, gy=gy)

cols = style(:x, :y);
grp = style(layout_x=:gx, layout_y=:gy)
scat = spec(Scatter)
pipeline = cols * scat
aog = data(df) * grp * pipeline
```

layout

Now faceting, in the ggplot sense.
facet

```julia
scene, layout = MakieLayout.layoutscene(fontsize=35, font="CMU Serif roman", resolution=round.(Int, (2 * 600, 2 * 540)))
layout.alignmode = Outside(0, 25, 0, 5)

AlgebraOfGraphics.layoutplot!(scene, layout, aog)
AbstractPlotting.save(joinpath("figures/layout.pdf"), scene)

x_facet_titles = sort!(unique(gx))
y_facet_titles = sort!(unique(gy))

grd_lay = layout.content[1].content

# remove xlabel and ylabel
for i in 1:length(grd_lay.content)
  ax = grd_lay.content[i].content
  ax.xlabelvisible[] = length(x_facet_titles) == 0
  ax.ylabelvisible[] = length(y_facet_titles) == 0
end

# add categories of x_facet variable and linked x_label
if length(x_facet_titles) > 0
  lax11 = grd_lay.content[1].content
  n = size(grd_lay)[1]
  for (i,t) in enumerate(x_facet_titles)
    text = LText(scene, t)
    text.tellheight = true
    text.tellwidth = false
    grd_lay[1,i,Top()] = LRect(scene, color = RGBAf0(0,0,0,0.2), strokevisible=false) 
    grd_lay[1,i,Top()] = text
  end
  
  xlabel_facet = LText(scene, string(lax11.xlabel[]))
  xlabel_facet.tellheight = true
  xlabel_facet.tellwidth = false
  grd_lay[end+1,:] = xlabel_facet
end

# add categories of y_facet variable and linked y_label
if length(y_facet_titles) > 0
  lax11 = grd_lay.content[1].content
  n = size(grd_lay)[2]
  for (i,t) in enumerate(y_facet_titles)
    text = LText(scene, t, rotation=π/2)
    text.tellheight = false
    text.tellwidth = true
    grd_lay[i,n, Right()] = LRect(scene, color = RGBAf0(0,0,0,0.2), strokevisible=false) 
    grd_lay[i,n, Right()] = text
  end
  
  ylabel_facet = LText(scene, string(lax11.ylabel[]), rotation = π/2)
  ylabel_facet.tellheight = false
  ylabel_facet.tellwidth = true
  grd_lay[1:end-1,0] = ylabel_facet
end
```

greimel:
Fixes #42

The case where layouting is done with multiple x and y columns (as in the tutorial) is unchanged.

```julia
iris = dataset("datasets", "iris")
cols = style([:SepalLength, :SepalWidth], [:PetalLength :PetalWidth])
grp = style(layout_x = dims(1), layout_y = dims(2), color = :Species)
geom = spec(Scatter) + linear
aog2 = data(iris) * cols * grp * geom
aog2 |> draw
```

layout2

The case where faceting is done with a categorical variable is improved: Note the shared x and y labels and the labels on top and to the right. 

```julia
mpg = dataset("ggplot2", "mpg")
mpg[!,:grpx] .= rand.(Ref(["a", "b", "c"]))
mpg[!,:grpy] .= rand.(Ref(["aa", "bb"]))

cols = style(:Displ, :Hwy)
grp = style(color = :Cyl => categorical, layout_x = :grpx, layout_y = :grpy)
scat = spec(Scatter)
pipeline = cols * scat
aog = data(mpg) * grp * pipeline
aog |> draw
```

layout1

Since I don't know to internal well enough, this could probably be implemented more elegantly. Let me know what I can do to improve the code.

Edit: Things to improve
* Should we also link all the x-axis when faceting by a categorical variable?
* Placement of legend is not nice


greimel:
I now use protrusions for spanned x and y labels so that the legend is correctly centered.

layout3

I would like to acknowledge help by @jkrumbiegel on Slack. He suggested to use protrusions for the grey labels a while ago and showed me how to use them + clever padding for the spanned labels.

src/makielayout_integration.jl
```diff
@@ -1,11 +1,11 @@
...
-                   linkxaxes!,
-                   linkyaxes!,
-                   hidexdecorations!,
-                   hideydecorations!,
-                   MakieLayout
+                   linkxaxes!, linkyaxes!,
+                   hidexdecorations!, hideydecorations!,
+                   MakieLayout,
+                   Top, Bottom, Left, Right,
+                   RGBAf0, lift, @lift, Node
```
greimel, line 8: where do these actually come from? is it ok to use them from MakieLayout?

jkrumbiegel: these are from MakieLayout

```
hidexdecorations!, hideydecorations!,
MakieLayout,
Top, Bottom, Left, Right,
```

`RGBAf0` is from Colors

and the others are from AbstractPlotting

asinghvi17: Top, Bottom, Left, Right are deprecated, I think.  RGBAf0 is from AbstractPlotting (shortcut for RGBA{Float32} from Colors) and the lift stuff is from AbstractPlotting also.

jkrumbiegel: > Top, Bottom, Left, Right are deprecated, I think
I didn't deprecate them, did someone else? Why would they be

src/makielayout_integration.jl
```diff
@@ -92,6 +105,72 @@ function layoutplot!(scene, layout, ts::Algebraic)
             @warn "Automated legend was not possible due to $e"
         end
     end
+    
+    ax1 = axs[end,1]
```
greimel, line 109: Due to my use of fake data, I didn't realize that I got the spanned x and y labels wrong. This is now fixed.

Now I take them from one `ax1`.

It would probably be preferable to take them directly from `split(trace.value)`?

src/makielayout_integration.jl
```diff
@@ -40,6 +45,13 @@ function set_defaults!(attrs::Attributes)
     get!(attrs, :markersize, Observable(8px))
 end
 
+pkeys(aog) = aog.dict.vals[1].context.pkeys
```
piever, line 48: Maybe better to specify the type of `aog`, as this seems very specific. `aog.dict.vals[1]` should probably be `first(values(aog))`.  Also, it'd probably be better to introduce `pkeys(c::AbstractContext)`, because the data context has a `pkeys` field, but other contexts don't, so `pkeys(::AbstractContext) = nothing` should probably be the general fallback.

src/makielayout_integration.jl
```diff
@@ -92,6 +105,72 @@ function layoutplot!(scene, layout, ts::Algebraic)
...
+            facetlayout[i, end, Right()] = text
+        end
+    
+        # Shared ylabel
+        group_left_protrusion = lift(
+            (xs...) -> maximum(y -> y.left, xs),
+            (MakieLayout.protrusionsobservable(ax) for ax in axs[:, 1])...
+        )
+    
+        pady = Node(10.0)
```
piever, line 164: Same as above on `padx`.

src/makielayout_integration.jl
```diff
@@ -92,6 +105,72 @@ function layoutplot!(scene, layout, ts::Algebraic)
...
+            facetlayout[1, i, Top()] = text
+        end
+    
+        # Shared xlabel
+        group_bottom_protrusion = lift(
+            (xs...) -> maximum(y -> y.bottom, xs),
+            (MakieLayout.protrusionsobservable(ax) for ax in axs[end, :])...
+        )
+    
+        padx = Node(10.0)
```
piever, line 137: Do we need this here? Somehow, I imagined that the defaults of MakieLayout should also be our defaults, or is there something MakieLayout is not doing automatically?

src/makielayout_integration.jl
piever: Did you experiment with the scenario where not all pairwise group combinations exist? That is, it's not a perfect grid but some plots are missing. If we want to support that, we'd need to make sure that `axs[end, 1]` does the right thing there.

src/makielayout_integration.jl
piever: Somehow, this feels a bit fragile, because `layoutplot!` accepts any `Algebraic`. Could you do it somewhere during the pipeline, where it is more clear what objects one is dealing with? For example, somewhere around here: https://github.com/JuliaPlots/AlgebraOfGraphics.jl/blob/master/src/makielayout_integration.jl#L49

piever: Actually never mind, this can probably stay as is, I'm working on a simplification of the system, so this will have to change in the near future anyways.

src/makielayout_integration.jl
greimel: @jkrumbiegel, would you mind helping us out? I basically copied your code from slack :-D 

Are similar the paddings between regular axis labels and tick labels similarly hard coded in MakieLayout? Or could extract the padding from the given plot and and use it here?

jkrumbiegel: well you could of course just use one of the padding attributes of one of the LAxis instances, it's the same thing but if you have a spanning label, this parameter doesn't belong to one of the spanned axes... I just wanted to show in the example how you can correctly position the label, it's your decision how you want to expose that parameter :)

greimel: thanks!

greimel: By default, the padding should be identical for all panels, right? (The axis labels are always nicely aligned). That is, I can just take the attribute from any panel.

jkrumbiegel: I'm not sure if I understand what you want to know. Normally, an axis has an x and y label and there is a padding value for those against the tick labels. What you're doing here is not using those labels at all, but simply sticking another label across the outer axes and padding it so it looks good. The extra padding value that you need can come from anywhere, but why would you reuse it from one of the axes? Then the user would have to go to that axis and change the padding value there. I think it makes most sense to make these values keyword arguments of the AoG function that creates the grid. I don't know if the AoG data structure in the end has a space for storing this attribute, though.

greimel: Thanks, I think I misunderstood which distance the padding measures.

PR #47, opened 2020-07-08
add Compat for AbstractPlotting

greimel:
v0.12 includes MakieLayout as submodule, but is not yet usable with CairoMakie.

In the meantime I propose to bind the version to 0.11

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,6 +17,7 @@ StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
+AbstractPlotting = "0.11"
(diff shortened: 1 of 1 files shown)


piever:
Sounds good!

asinghvi17:
@greimel why does CairoMakie not work?

greimel:
probably just because of the version bound. I hadn't realized that when I upgraded AbstractPlotting it downgraded CairoMakie to a very old version. Sorry for the confusion.

It's still true that code in this package needs to be updated for AbstractPlotting 0.12, and this should be done, after the version bound in CairoMakie is adjusted.

piever:
> why does CairoMakie not work?

CairoMakie has GeoMakie as a test dependency, and GeoMakie has not added support for AbstractPlotting 0.12.

PR #54, opened 2020-07-17
Faceting: robustify spanned labels

greimel:
Fixes #50.
``` julia
Pkg.pkg"add https://github.com/JuliaPlots/CairoMakie.jl#pv/update"

using RDatasets: dataset
using AlgebraOfGraphics, AbstractPlotting, CairoMakie
using Underscores, DataFrames

# data
mpg = dataset("ggplot2", "mpg")
mpg[!,:grpx] .= rand.(Ref(["a", "b", "c"]))
mpg[!,:grpy] .= rand.(Ref(["aa", "bb"]))

dta1 = mpg |> data

# remove data to get empty panels
dta2 = @_ mpg |> 
    filter(!((_.grpx == levels(__.grpx)[2]) & (_.grpy == levels(__.grpy)[1])), __) |>
    filter(!((_.grpx == levels(__.grpx)[1]) & (_.grpy == levels(__.grpy)[2])), __) |>
    data

# specify plots    
cols1 = style(:Displ, :Hwy)
grp1 = style(color = :Cyl => categorical, layout_x = :grpx, layout_y = :grpy)
geom = spec(Scatter)

aog1 = dta1 * grp1 * cols1 * geom
aog2 = dta2 * grp1 * cols1 * geom

scn1 = aog1 |> draw # just fine
scn2 = aog2 |> draw # no labels before this PR
```

Before:

facet2-before

After:

facet2-after

PR #58, opened 2020-07-18
fix axis labels of single row and single column plots

greimel:
This fixes a bug introduced in #54: for plots without layout_{x,y} the {x,y} label is not shown.

This is fixed now:

Case 1:
facet0

Case 2:
facet1

Case 3:
facet2

Case 4:
facetx

Case 5:
facety

### DIFF
--- a/src/draw.jl
+++ b/src/draw.jl
@@ -160,7 +160,7 @@ function spannable_xy_labels(layout)
         xlab = nothing
     end
     
-    # if layout has multiple row, check if xlabel is spannable
+    # if layout has multiple rows, check if xlabel is spannable
     if size(layout)[1] > 1
         unique_y_labs = unique(labs.y[.! labs.empty])
         ylab = length(unique_y_labs) == 1 ? only(unique_y_labs) : nothing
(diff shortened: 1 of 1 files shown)


src/draw.jl
```diff
@@ -152,12 +152,22 @@ function spannable_xy_labels(layout)
...
-    ylab = length(unique_y_labs) == 1 ? only(unique_y_labs) : nothing
+    # if layout has multiple columns, check if xlabel is spannable
+    if size(layout)[2] > 1
+        unique_x_labs = unique(labs.x[.! labs.empty])
+        xlab = length(unique_x_labs) == 1 ? only(unique_x_labs) : nothing
+    else
+        xlab = nothing
+    end
     
+    # if layout has multiple row, check if xlabel is spannable
```
piever, line 163:
```suggestion
    # if layout has multiple rows, check if xlabel is spannable
```

piever:
Ups, yes, that's true, I've just noticed a bunch of plots had lost their labels.

greimel:
Some visual regression tests would be nice ;-)

PR #59, opened 2020-07-18
Padding of spanned axis labels

greimel:
Instead of hard-coding the padding, the retrieves the padding from the individual panels, checks that they are all the same and use them.

Before:
facetx

After:
facetx-after

I don't really know how to handle observables. If you want that the padding of the spanned labels listens to the padding of the individual panels, feel free to close this and use this commit in a separate PR.

### DIFF
--- a/src/draw.jl
+++ b/src/draw.jl
@@ -160,7 +160,7 @@ function spannable_xy_labels(layout)
         xlab = nothing
     end
     
-    # if layout has multiple row, check if xlabel is spannable
+    # if layout has multiple rows, check if xlabel is spannable
     if size(layout)[1] > 1
         unique_y_labs = unique(labs.y[.! labs.empty])
         ylab = length(unique_y_labs) == 1 ? only(unique_y_labs) : nothing
@@ -180,4 +180,3 @@ end
 layoutplot(; kwargs...) = t -> layoutplot(t; kwargs...)
 
 draw(args...; kwargs...) = layoutplot(args...; kwargs...)
-
--- a/src/draw.jl
+++ b/src/draw.jl
@@ -152,12 +152,22 @@ function spannable_xy_labels(layout)
         (x = ax.xlabel[], y = ax.ylabel[], empty = isemptyax(ax))
     end |> StructArray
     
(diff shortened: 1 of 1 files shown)


src/draw.jl
```diff
@@ -149,16 +149,41 @@ end
...
+    else
+        xlab = nothing
+        xpad = nothing
+    end
     
-    (x = xlab, y = ylab)
+    # if layout has multiple rows, check if xlabel is spannable
+    if size(layout)[1] > 1
+        unique_y_labs = unique(labs.y[.! labs.empty])
+        ylab = length(unique_y_labs) == 1 ? only(unique_y_labs) : nothing
```
piever, line 174: Maybe refactor this as a utility function? It is used four times (padding and label for x and y axis).

src/draw.jl
```diff
@@ -149,16 +149,41 @@ end
...
+            ypad = only(unique_y_pads)
+        else
+            ypad = nothing
+        end  
+    else
+        ylab = nothing
+        ypad = nothing
+    end
+        
+    (x = (lab=xlab, pad=xpad), y = (lab=ylab, pad=ypad))
```
piever, line 186: I think it's clearer without abbreviations, that is `(label=xlab, padding=xpad)`.

src/draw.jl
```diff
@@ -149,16 +149,41 @@ end
 function spannable_xy_labels(layout)
     labs = map(layout.content) do _ax
         ax = _ax.content
-        (x = ax.xlabel[], y = ax.ylabel[], empty = isemptyax(ax))
+        (x = ax.xlabel[], y = ax.ylabel[],
```
piever, line 152: This doesn't really work for observables, because once you do `ax.xlabel[]` (note the square brackets), it is no longer responsive. You should take `ax.xlabel` as is, and pass it around (lifting when necessary). Same for `ax.xlabelpadding` (especially because afterwards you wrap it in a new `Node`, which you shouldn't need to). So basically, instead of

```julia
padx = Node(ax.xlabelpadding[]) # no longer responds to changes in `ax.xlabelpadding`
```

you simply want

```julia
padx = ax.xlabelpadding
```

PR #68, opened 2020-07-23
Remove duplicated legends when grouping variable is used more than once

greimel:
Currently, if a grouping variable is used twice, this will produce two identical legends.

```julia
using RDatasets: dataset
using AlgebraOfGraphics, AbstractPlotting, CairoMakie
mpg = dataset("ggplot2", "mpg");
cols = style(:Displ, :Hwy);
grp = style(color = :Cyl => categorical, marker=:Cyl => categorical);
scat = spec(Scatter, markersize=10px)
pipeline = cols * scat
scn = data(mpg) * grp * pipeline |> draw
```

before:
before

with this PR this doesn't happen any longer
after

probably the implementation is overkill. It seems (but I cannot prove) that the for a given grouping variable the legends are always identical. In that case a simple filtering would suffice.

### DIFF
--- a/src/legend.jl
+++ b/src/legend.jl
@@ -34,6 +34,7 @@ function add_entry!(legend::Legend, entry::String; title::String="")
 end
 
 function create_legend(scene, legend::Legend)
+    legend = remove_duplicates(legend)
     sections = legend.sections
     MakieLayout.LLegend(
         scene,
@@ -42,3 +43,28 @@ function create_legend(scene, legend::Legend)
         getproperty.(sections, :title)
     )
 end
+
+function remove_duplicates(legend)
+    sections = legend.sections
+    titles = getproperty.(sections, :title)
+    # check if there are duplicate titles
+    unique_inds = unique_indices(titles)
+    has_duplicates = length(unique_inds) < length(titles)
+    # if so: remove duplicates, generate new names
+    if has_duplicates
+        sections_new = sections[unique_inds]
+    
+        names_new = map(unique_inds) do i
+            title = titles[i]
+            duplicate_inds = titles .== title
+            # ["var1", "var2"] becomes ["var1_var2"]
+            name = reduce((s,t) -> s * "_" * t, legend.names[duplicate_inds])
+        end
+        
+        return Legend(names_new, sections_new)
+    else
+        return legend
+    end
+end
+
+unique_indices(x) = findfirst.(isequal.(unique(x)), [x])
--- a/src/legend.jl
+++ b/src/legend.jl
@@ -67,8 +67,6 @@ function remove_duplicates(legend)
     end
 end
 
-unique_indices(x) = findfirst.(isequal.(unique(x)), [x])
-
 function unique_indices(x; keep)
     first_inds = indexin(x, x)
     inds_keep = findall(==(keep), x)
--- a/src/legend.jl
+++ b/src/legend.jl
@@ -70,8 +70,7 @@ end
 unique_indices(x) = findfirst.(isequal.(unique(x)), [x])
 
 function unique_indices(x; keep)
-    unique_inds = unique_indices(x)
+    first_inds = indexin(x, x)
     inds_keep = findall(==(keep), x)
     sort!(unique!([unique_inds; inds_keep]))
 end
-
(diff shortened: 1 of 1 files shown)


piever:
Thanks!

> It seems (but I cannot prove) that the for a given grouping variable the legends are always identical. In that case a simple filtering would suffice.

That's correct. The only exception is the case `col => somefunction`, so in principle one could pass the same column twice with different functions, and it would give two different columns with the same name (but that's a separate issue, I've opened #69). 

This may change in the future. I was thinking that the "marker" legend should be in black, but I guess we'll worry about that when that happens. Could you change this to just use the filtering?

One more thing is that there are contexts where the legend titles are not inferred and default to `" "` (a single space). I think it'd be good to only "deduplicate" the legend if there are no blank (i.e., `== " "`) titles, as there duplicated titles would just be a failure to do the automatic thing. (In `add_entry!` the default is `title=""`, but that's wrong, it should actually be `" "`, the empty string creates problems with MakieLayout.)

greimel:
do you have a test case where title are not inferred?

greimel:
anyways, this should work, and is ready for review

piever:
> do you have a test case where title are not inferred?

Yes, the [last examples in the tutorial](http://juliaplots.org/AlgebraOfGraphics.jl/dev/generated/tutorial/#Non-tabular-style-(slicing-context)-1). You can simply try building the docs locally, with `julia --project=docs` and then `include("docs/make.jl")`, and see if everything looks reasonable.

src/legend.jl
```diff
@@ -34,11 +34,44 @@ function add_entry!(legend::Legend, entry::String; title::String="")
...
+    has_duplicates = length(unique_inds) < length(titles)
+    # if so: remove duplicates, generate new names
+    if has_duplicates
+        sections_new = sections[unique_inds]
+    
+        names_new = map(unique_inds) do i
+            title = titles[i]
+            duplicate_inds = titles .== title
+            # ["var1", "var2"] becomes ["var1_var2"]
+            name = reduce((s,t) -> s * "_" * t, legend.names[duplicate_inds])
```
piever, line 61: I guess `join(legend.names[duplicate_inds], '_')` is the standard way of writing this, but I'm not completely sure we need it (`legend.names` is not used in making the graphical legend I think). I think just

```julia
names_new = legend.names[unique_inds]
```

is fine.

src/legend.jl
```diff
@@ -34,11 +34,44 @@ function add_entry!(legend::Legend, entry::String; title::String="")
...
+        return Legend(names_new, sections_new)
+    else
+        return legend
+    end
+end
+
+unique_indices(x) = findfirst.(isequal.(unique(x)), [x])
+
+function unique_indices(x; keep)
+    unique_inds = unique_indices(x)
```
piever, line 73:
```suggestion
    first_inds = indexin(x, x)
```

src/legend.jl
```diff
@@ -34,11 +34,44 @@ function add_entry!(legend::Legend, entry::String; title::String="")
...
+        end
+        
+        return Legend(names_new, sections_new)
+    else
+        return legend
+    end
+end
+
+unique_indices(x) = findfirst.(isequal.(unique(x)), [x])
+
```
piever, line 71:
```suggestion
```

src/legend.jl
```diff
@@ -34,11 +34,44 @@ function add_entry!(legend::Legend, entry::String; title::String="")
...
+function unique_indices(x; keep)
+    unique_inds = unique_indices(x)
+    inds_keep = findall(==(keep), x)
+    sort!(unique!([unique_inds; inds_keep]))
```
piever, line 75:
```suggestion
    sort!(union!(inds_keep, first_inds))
```

greimel:
I built the docs locally and everything looks fine.

There was some warning that `histogram` is not defined. But that's unrelated.

src/legend.jl
```diff
@@ -34,11 +34,35 @@ function add_entry!(legend::Legend, entry::String; title::String="")
...
+    sections = legend.sections
+    titles = getproperty.(sections, :title)
+    # check if there are duplicate titles
+    unique_inds = unique_indices(titles; keep = " ")
+    has_duplicates = length(unique_inds) < length(titles)
+    # if so: remove duplicates, generate new names
+    if has_duplicates
+        sections_new = sections[unique_inds]
+        names_new = legend.names[unique_inds]
+
```
piever, line 57:
```suggestion
```

PR #77, opened 2020-08-16
pad facet text, add rect stroke and brighten color for cleaner look

jkrumbiegel:
with the changes I just made to the axis spine and rect alignments in abstractplotting (still needs to be merged and tagged) I think this improves the look. the text padding is good so that larger letters like g y etc. don't touch the borders

### DIFF
--- a/src/draw.jl
+++ b/src/draw.jl
@@ -37,9 +37,11 @@ function add_facet_labels!(scene, axs, layout_levels;
     for i in eachindex(lxl)
         pos = axis == :x ? (1, i, Top()) : (i, Nx, Right())
         facetlayout[pos...] = LRect(
-            scene, color = RGBAf0(0, 0, 0, 0.2), strokevisible=false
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
grafik

this is what it will look like witht the AbstractPlotting changes

jkrumbiegel:
grafik
vs now, see the g on the right label

piever:
Thanks, looks much nicer!

One question, on the generated docs (with CairoMakie) it seems that the text area linewidth is much thicker than the axis linewidth (see [this section](http://juliaplots.org/AlgebraOfGraphics.jl/previews/PR77/generated/tutorial/#Layout-1)). Is it a CairoMakie issue, or will this also be fixed with your AbstractPlotting changes?

jkrumbiegel:
Yeah the LRect default strokewidth was 2 for some reason, I set it to 1 like everything else in the latest change :)

PR #78, opened 2020-08-17
Hide empty legend titles

jkrumbiegel:
LLegend shows an empty row for an empty string, so we pass `nothing` for empty titles instead


piever:
Ah, didn't know about that, thanks!

Issue #82, opened 2020-08-18
Legend: colorbars for continuous variables are missing


greimel:
![col](https://user-images.githubusercontent.com/6280307/90514863-0c710700-e162-11ea-98b7-aa6f49c1890f.png)

piever:
Fixed by #164

PR #84, opened 2020-08-18
[WIP] Add colorbar

greimel:
```julia
using RDatasets: dataset
using AlgebraOfGraphics, AbstractPlotting, CairoMakie
mpg = dataset("ggplot2", "mpg")
cols = style(:Displ, :Hwy)
grp = style(color = :Cyl => categorical)
mpg[!,:grp_cyl] = mpg.Cyl .> 6
mpg[!,:col] = 3 .* rand.() .+ 2 .* mpg.grp_cyl 

scn = data(mpg) * style(:Displ, :Hwy, color = :col, layout_x = :grp_cyl => categorical) * spec(Scatter) |> draw
```

colormap

The function in colorbar.jl should probably go to AbstractPlotting.MakieLayout when it's done. (Should probably work with observables)

EDIT: Thinking more about it, a function `linkcolorranges!` (similar to `linkxaxes!`) would probably be the better way to do it. What do you think, @jkrumbiegel?

### DIFF
--- a/src/AlgebraOfGraphics.jl
+++ b/src/AlgebraOfGraphics.jl
@@ -49,6 +49,7 @@ include("analysis/frequency.jl")
 include("analysis/histogram.jl")
 include("legend.jl")
 include("dodge.jl")
+include("colorbar.jl")
 include("draw.jl")
 
 end # module
--- a/src/colorbar.jl
+++ b/src/colorbar.jl
@@ -1,4 +1,6 @@
-function AbstractPlotting.MakieLayout.LColorbar(parent::Scene, plots::Vector{<: AbstractPlot}; kwargs...)
+import AbstractPlotting.MakieLayout.LColorbar
+
+function AbstractPlotting.MakieLayout.LColorbar(parent, plots::Vector{<: AbstractPlot}; kwargs...)
     colormap = plots[1].colormap
     # compute colorrange
     min = minimum(p.colorrange[][1] for p in plots)
--- a/src/colorbar.jl
+++ b/src/colorbar.jl
@@ -1,0 +1,16 @@
+function AbstractPlotting.MakieLayout.LColorbar(parent::Scene, plots::Vector{<: AbstractPlot}; kwargs...)
+    colormap = plots[1].colormap
+    # compute colorrange
+    min = minimum(p.colorrange[][1] for p in plots)
+    max = maximum(p.colorrange[][2] for p in plots)
+    colorrange = (min, max)
+     
+    for p in plots
+        p.colorrange = colorrange
+    end
+    
+    LColorbar(parent,
+        colormap = plots[1].colormap,
+        limits = colorrange;
+        kwargs...) 
+end
--- a/src/draw.jl
+++ b/src/draw.jl
@@ -165,17 +165,24 @@ function layoutplot!(scene, layout, ts::ElementOrList)
             end
         end
     end
+    
+    legend_layout = layout[1, end+1] = GridLayout(tellheight = false)
+    
     if !isempty(legend.sections)
         try
-            layout[1, end+1] = create_legend(scene, legend)
+            leg = legend_layout[end+1, 1] = create_legend(scene, legend)
+            leg.framevisible[] = false
         catch e
             @warn "Automated legend was not possible due to $e"
         end
     end
     if length(for_colormap) > 0
         T = typeof(for_colormap[1])
-        layout[1,end+1] = MakieLayout.LColorbar(scene, T[for_colormap...], width=30, label = string(colorname))
+        cbar = legend_layout[end+1, 1] = MakieLayout.LColorbar(scene, T[for_colormap...], width=30, height=120)
+        legend_layout[end, 1, Top()] = LText(scene, string(colorname), padding = (10,10,10,10))
     end
+    
+    trim!(legend_layout)
     
     layout_x_levels = get(level_dict, :layout_x, nothing)
     layout_y_levels = get(level_dict, :layout_y, nothing)
--- a/src/draw.jl
+++ b/src/draw.jl
@@ -109,6 +109,7 @@ function layoutplot!(scene, layout, ts::ElementOrList)
     hideydecorations!.(axs[:, 2:end], grid = false)
     
     for_colormap = []
+    colorname = nothing
     
     legend = Legend()
     level_dict = Dict{Symbol, Any}()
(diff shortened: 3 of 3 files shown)


greimel:
fixes #82

greimel:
now with label

colorbar

jkrumbiegel:
I would switch the logic around. Lift one colorrange from each plot, lift one max range out of those, pass that to the plots and the colorbar.

greimel:
I thought that this might be a problem, because one "overwrites" the colorranges in the plots. so that they don't update anymore when the inputs change?

would one have to use the z vector directly, use all of the z vectors to get the master-colorrange and feed that into the plot and colorbar?

jkrumbiegel:
It depends a bit how much "updatability" you need / want in AlgebraOfGraphics. Updating data always complicates things, especially if you get into layouts, legends, etc. which have to change given different data.

> one "overwrites" the colorranges in the plots. so that they don't update anymore when the inputs change?

That just depends on which way you choose to implement it. For me, it's usually easier to construct the Node up front, and then pass it to the object at creation. So in your case, a master colorrange node.

greimel:
if that means that you are opposed to a function `linkcolorranges!` in AbstractPlotting, I can hack something together here.

If you think that such a function would make sense, I'll think a bit more about it.

jkrumbiegel:
I think that would need to be implemented as a low-level feature in AbstractPlotting, so that every plot object in its color sampler can store a list of linked plot objects that also have color sampler. Otherwise you get really messy observables quite soon if you all link them up

piever:
Is there a specific consensus whether the colorrange should be the same across plots? As far as I understand from ggplot2, it must be the same across layers, but can be allowed to vary (not by default though) across facets.

Note that there is some scale machinery to compute extrema globally [here](https://github.com/JuliaPlots/AlgebraOfGraphics.jl/blob/master/src/scales.jl#L42), which needs a revamp for better continuous scale support.

jkrumbiegel:
I think usually, if there's just one variable that is translated into a color, you get only one colorbar and therefore one color range, even if another variable splits across facets. But I guess there could be a need for facets where each facet has its own color bar, so I guess a switch might be helpful to distribute one colorbar per facet. But the single colorbar thing is definitely the most common scenario I'd say

greimel:
@piever, you are getting at a bigger issue here. Should scales be linked by default?

So far, axes are linked, whenever it allows to save space (drop x-labels for stacked plots and and y-labels for dodged plots). So, linking colorbars is consistent with this design.

Similar to #85, we need to find a way to pass kwargs like `link_x`, `link_y`, `link_color`, etc

I think, all should be linked by default, with the option to unlink.

jkrumbiegel:
wouldn't it be a good solution to go back to types for AoG specific options like these?

Instead of 
```julia
style(:width, :length, layout_x = :Species, link_x = true, ...)
```
we could do
```julia
style(:width, :length, LayoutX(:species, link = true, label = :shared))
```

greimel:
yes, that would be nice. 

Where would you put `link_color`? Is it a problem that it doesn't have meaning in `Color()` unless `LayoutX()` or `LayoutY()` are used?

jkrumbiegel:
> Where would you put link_color?

Hm good point, maybe it's better not to separate these three things then. What about one of these

```julia
style(:width, :length, Facets(:species, link_x = true, link_color = true)

style(:width, :length, Facets(:species, link = [:x, :color, :markersize, ...])
```

I wonder if it's better in that case to have only `Facets` with multiple dispatch for x, y or x & y (how to separate x or y with one argument though?) or `FacetsX`, `FacetsY`, `FacetsXY` and `FacetWrap`

greimel:
Improved the alignment in the presence of colorbar and other legends

```julia
using RDatasets: dataset
using AlgebraOfGraphics, AbstractPlotting, GLMakie
using AlgebraOfGraphics.MakieLayout

mpg = dataset("ggplot2", "mpg")
cols = style(:Displ, :Hwy)
grp = style(color = :Cyl => categorical)
mpg[!,:Group] = mpg.Cyl .> 6
mpg[!,:Color] = 3 .* rand.() .+ 2 .* mpg.grp_cyl 

aog2 = data(mpg) * style(:Displ, :Hwy, layout_x = :Drv, marker = :Group => categorical, color = :Color) * spec(Scatter)

scene, layout = layoutscene()
 AlgebraOfGraphics.layoutplot!(scene, layout, aog2)

```
Screenshot 2020-09-25 at 16 24 05

greimel:
and here is some code to place the legend at the top

Screenshot 2020-09-25 at 16 44 43

```julia
scene, layout = layoutscene()
 AlgebraOfGraphics.layoutplot!(scene, layout, aog2)

 position = :top
 leg_lay = contents(layout[1,2])[1]
 leg = contents(leg_lay[1,1])[1]
 bar = contents(leg_lay[2,1])[1]
 bar_title = contents(leg_lay[2,1, Top()])[1]

 leg.orientation[]   = position == :right ? :vertical : :horizontal
 leg.titleposition[] = position == :right ? :top : :left 
 leg.tellheight[] = true
 leg.tellwidth[] = true
 
 bar.vertical[] = position == :right
 if position == :right
   bar.height[] = 120
   bar.width[] = 20 #Relative(0.5)
   bar.ticksize[] = 10
   bar.ticklabelpad[] = 5
   bar.ticklabelsize[] = 20
 else
   bar.width[] = 120
   bar.height[] = 20
   bar.ticklabelalign = (:center, :bottom)
   bar.ticksize[] = 5
   bar.ticklabelpad[] = 1.5
   bar.ticklabelsize[] = 15
 end

 if position == :right
    leg_lay.tellheight[] = false
    leg_lay.tellwidth[] = true
    leg_lay[2,1] = bar
    leg_lay[2,1,Top()] = bar_title
    layout[1,2] = leg_lay
 elseif position == :top
   leg_lay.tellwidth[] = false
   leg_lay.tellheight[] = true
    leg_lay[1,2] = bar
    leg_lay[1,2,Left()] = bar_title
    layout[0,1] = leg_lay
 end
 trim!(leg_lay)
 trim!(layout)

 scene
```

greimel:
In case of a vertical colorbar, it should be moved to the left, so that it is approximately aligned with the markers. This should be achievable by wrapping it in a GridLayout.

greimel:
implemented as part of https://github.com/piever/SplitApplyPlot.jl/  refactor.

PR #92, opened 2020-09-29
use NamedDims unname to remove names

oxinabox:
`unname` is the operation that semantically does this:
See:
https://github.com/invenia/NamedDims.jl/blob/942bb283c62dd64c0ca484bda19c1bb832f268fa/src/name_operations.jl#L45-L46

Under the hood it just calls `parent` but seem more semantic to use `unname`.
Potentially could just replace all uses of `strip_names` with `unname`


piever:
Thanks! Happy to merge when CI passes (looks like `unname` is not being imported from NamedDims).

Issue #104, opened 2020-10-22
Axis labels don't show up


ericphanson:
I've been trying to follow the [first example in the 0.1.2 tutorial](http://juliaplots.org/AlgebraOfGraphics.jl/stable/generated/tutorial/#Working-with-tables-1), but the axis labels don't show up for me:

![scatter](https://user-images.githubusercontent.com/5846501/96862996-82496580-1466-11eb-9068-d9b1c099b9fe.png)

(with SVG or PNG, but Github doesn't let me upload the SVG). This was produced by cloning the repo, checking out the 0.1.2 tag, starting Julia 1.4.2 with `julia --project=docs`, doing `] dev .` to dev AlgebraOfGraphics, then running the tutorial code.

I think it must be due to different versions of dependencies and transitive dependencies, but I've been having a lot of trouble tracking it down. There is some weird stuff, like the 0.1.2 tag of AlgebraOfGraphics has [AbstractPlotting compat'd to 0.12](https://github.com/JuliaPlots/AlgebraOfGraphics.jl/blob/v0.1.2/Project.toml) but the [checked-in docs manifest has Abstractplotting at 0.9.27](https://github.com/JuliaPlots/AlgebraOfGraphics.jl/blob/45b1bb9b2166ac8fa07a89d01d08a317a38029ab/docs/Manifest.toml#L13). I think what happens is that in the [`Pkg.develop` step of the Travis docs build](https://github.com/JuliaPlots/AlgebraOfGraphics.jl/blob/45b1bb9b2166ac8fa07a89d01d08a317a38029ab/.travis.yml#L27) the package resolver chooses new versions of all the dependencies and updates its local copy of the Manifest, then generates the docs, and that Manifest is never checked in (which also happens with my local `] dev .` step). So I'm not actually sure how to repro the docs environment to try to get axis labels. However, if you want to repro my environment that does not have axis labels, here's the Manifest and Project I get after `] dev .` in the docs project:

[no-axis-label-env.zip](https://github.com/JuliaPlots/AlgebraOfGraphics.jl/files/5422210/no-axis-label-env.zip)

By the way, I think the reproducibility problem for the docs build can be solved by checking in a Manifest with AlgebraOfGraphics already dev'd by a relative path, and then not `dev`ing it during the docs build step, to not let the resolver alter the Manifest.

piever:
Turns out I was working around a hack in MakieLayout that was fixed in the main time, which in turn broke my hack...

[...]

PR #105, opened 2020-10-22
remove manifest from the docs

piever:
TODO: figure out regression on axis labels #104


ericphanson:
Ah good, it isn't just me 😄 http://juliaplots.org/AlgebraOfGraphics.jl/previews/PR105/generated/tutorial/

piever:
Yes, looks like there were some changes / regressions in AbstractPlotting. I'll try to fix this and make a new release. Thanks for spotting those!

I think it's best to not have the manifest, so it's easy to check with the preview of each PR that things are working with the latest versions of everything. (I haven't set up proper visual regression tests, as they can be a bit painful, so I essentially check out generated docs.)

PR #137, opened 2021-01-17
Rename types from AbstractPlotting.

adkabo:
AbstractPlotting renamed a bunch of things recently.
This change propagates those renames to AlgebraOfGraphics.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -18,7 +18,7 @@ StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
-AbstractPlotting = "0.12.17, 0.13, 0.14"
+AbstractPlotting = "0.15"
 CategoricalArrays = "0.8"
 GLM = "1"
 GeometryBasics = "0.2, 0.3"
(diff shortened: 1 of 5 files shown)


adkabo:
I've only tested a bit manually. It seems okay so far.

greimel:
the failure on 1.4 should be fixed by https://github.com/JuliaPlots/AbstractPlotting.jl/pull/601

@piever, the docs seem to be built for every PR. Do you know how to look at them?

piever:
Thanks for doing this!

I wouldn't worry about the julia 1.4 failure: if it's a problem with loading AbstractPlotting, it is up to AbstractPlotting to put the bound on the julia version (or to fix, if it's easy enough).

The docs are generated at addresses like http://juliaplots.org/AlgebraOfGraphics.jl/previews/PR95/ (with the appropriate PR number), but unfortunately it only works if the PR comes from a branch of the JuliaPlots AlgebraOfGraphics repo, not from a fork.
For users working from their fork, I'm afraid the only way is to build the docs locally.

@adkabo this actually looks good to me, as the failure is unrelated. If you mark the PR as ready for review, I think it can be merged.

adkabo:
Sure, marked as ready.

PR #139, opened 2021-01-20
Switch to Figure object

piever:
TODO:

- [x] Understand what happens when the facet grid is not "full"
- [x] Understand why the generated docs at http://juliaplots.org/AlgebraOfGraphics.jl/previews/PR139/generated/tutorial/ look catastrophic.

### DIFF
--- a/src/draw.jl
+++ b/src/draw.jl
@@ -99,8 +99,7 @@ function draw!(figure, ts::ElementOrList)
         # dodge may need to be done separately per each subplot
         Ndodge = max(Ndodge, rank(to_value(get(spec.options, :dodge, Ndodge))))
     end
-    # What is a better placeholder here, in case some plots are missing?
-    axs = Union{Axis, Nothing}[nothing for i in 1:Ny, j in 1:Nx]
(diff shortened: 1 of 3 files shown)


src/draw.jl
```diff
@@ -103,16 +99,8 @@ function layoutplot!(scene, layout, ts::ElementOrList)
...
-        linkxaxes!(axs[:, i]...)
-    end
-    for i in 1:Ny
-        linkyaxes!(axs[i, :]...)
-    end
-    hidexdecorations!.(axs[1:end-1, :], grid = false)
-    hideydecorations!.(axs[:, 2:end], grid = false)
-
+    # What is a better placeholder here, in case some plots are missing?
+    axs = Union{Axis, Nothing}[nothing for i in 1:Ny, j in 1:Nx]
```
greimel, line 103: ggplot displays axes even if there is no data to display. This works since all axes are linked by default.



Only for facet_wrap (which we don't have), it doesn't show empty axes.

piever: Yes, I've updated it to always start with an axis placeholder, I agree with you that it makes more sense.

PR #175, opened 2021-05-16
add choropleth

piever:
@visr this is my attempt of converting GeoInterface to GeometryBasics.

I assumed hat:

- `coords = GeoInterface.coordinates(shp)` returns a list, in which each entry correspond to a polygon
- `first(coords)` is in turn a list, the first entry being the exterior of the polygon, the remaing the interior
- `points = first(first(coords))` is a list of points, represented as a list of vectors of length 2.

Is that all correct?

I've added an example from an Antarctic shapefile to respect the penguin theme. If you know of a nice (possibly antarctic) shapefile (maybe a bit more lightweight, this is very detailed), I'm open to suggestions.

Plan is to move the conversion function inside of AlgebraOfGraphics (or maybe even AbstractPlotting) for simplicity.

Link to docs preview: http://juliaplots.org/AlgebraOfGraphics.jl/previews/PR175/generated/gallery/#Geographic-data


visr:
> Is that all correct?

That is indeed all correct! Quite the nesting isn't it?

I quite like this example, the picture looks great. It helps that the projection is already good, if it would be latitude, longitude, it would look terrible.

I'm not really aware of other Antarctic shapefiles. If you want I can very easily do a simplify on this one to reduce the detail a bit. We'd need to host the derived data somewhere though.

So the conversion code you created will handle the MultiPolygon case. If you also want to handle the Polygon case, you'd need to strip the outer map and just create a single GeometryBasics.Polygon. You can keep the two apart with a check like this:

```
GeoInterface.geotype(shp) == :MultiPolygon
GeoInterface.geotype(shp) == :Polygon
```

piever:
Thanks for the tip on the different supported `geotype`, I've changed the code to support both `Polygon` and `MultiPolygon`.

> I quite like this example, the picture looks great. It helps that the projection is already good, if it would be latitude, longitude, it would look terrible.
> 
> I'm not really aware of other Antarctic shapefiles. If you want I can very easily do a simplify on this one to reduce the detail a bit. We'd need to host the derived data somewhere though.

Thanks! But in the end I was worrying about nothing. The  "medium resolution" they offer is actually OK, the filesize is small and the resulting svg of the plot is also of a reasonable size, so I'll probably just stick with this one.

PR #178, opened 2021-05-18
clarify tutorial fit grouping

knuesel:
The penguin tutorial shows linear fits on female and male plots using

```julia
layers = linear() + mapping(col = :sex)
plt = penguin_bill * layers * mapping(color = :species)
draw(plt; axis)
```

image

The sex mapping is applied only to the scatter so the same fit is shown left and right. Is this intentional? This PR first shows a specific fit for each sex:

```julia
layers = linear() + mapping()
plt = penguin_bill * layers * mapping(color = :species, col = :sex)
draw(plt; axis)
```

image

then shows how to pool data from both sexes (i.e showing the original code).

### DIFF
--- a/docs/src/generated/penguins.jl
+++ b/docs/src/generated/penguins.jl
@@ -112,7 +112,7 @@ layers = linear() + mapping(col = :sex)
 plt = penguin_bill * layers * mapping(color = :species)
 draw(plt; axis)
 
-# See how both plots show the same fit, because the sex mapping is not applied
+# See how both plots show the same fit, because the `sex` mapping is not applied
 # to `linear()`. The following on the other hand produces a separate fit for
 # males and females:
 
--- a/docs/src/generated/penguins.jl
+++ b/docs/src/generated/penguins.jl
@@ -105,11 +105,19 @@ layers = linear() + mapping(marker = :sex)
 plt = penguin_bill * layers * mapping(color = :species)
 draw(plt; axis)
 
-# This plot is getting a little bit crowded. We could instead analyze female and
+# This plot is getting a little bit crowded. We could instead show female and
 # male penguins in separate subplots.
 
 layers = linear() + mapping(col = :sex)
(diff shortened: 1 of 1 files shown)


piever:
That's intentional, as this technique (different groupings for scatter and linear fit) was introduced in the previous paragraph. Indeed, in the previous plot, males and females have different scatters but the linear fit is shared.

If that's unclear, maybe it'd be better to just add a few words specifying that we keep using all members from each species for the fit.

knuesel:
Ah that makes sense. Happy to update the PR.

What about showing the other plot (one fit per sex) afterwards then, to make it clear? I think it's a subtle but important point. I love that AlgebraOfGraphics makes both options super easy, but that also means easy to get wrong if the user has only a superficial understanding of the algebra. People might not read carefully all the text in the tutorial... They might look at the image, interpret it as I did as different fits for different plots (probably the more common case?), copy-paste the code and get wrong results without realizing.

knuesel:
I have updated the PR to show what I have in mind. Do you prefer that I just remove the figure with independent fits?

docs/src/generated/penguins.jl
```diff
@@ -105,13 +105,21 @@ layers = linear() + mapping(marker = :sex)
...
 
-# This plot is getting a little bit crowded. We could instead analyze female and
+# This plot is getting a little bit crowded. We could instead show female and
 # male penguins in separate subplots.
 
 layers = linear() + mapping(col = :sex)
 plt = penguin_bill * layers * mapping(color = :species)
 draw(plt; axis)
 
+# See how both plots show the same fit, because the sex mapping is not applied
```
piever, line 115:
```suggestion
# See how both plots show the same fit, because the `sex `mapping is not applied
```

piever:
Thanks, that's a helpful change. I've just left a small comment.

knuesel:
Thanks, I've integrated the suggestion.

piever:
Thanks again for the PR!

PR #181, opened 2021-05-20
switch to Makie

piever:


AriMKatz:
Hi, what are the implications of this? Still backend agnostic ?

piever:
No, it works exclusively with Makie. It stopped being backend agnostic a while ago (I've updated the README), this PR just depends on Makie directly rather than on AbstractPlotting (where Makie's code used to live till not long ago).

PR #201, opened 2021-06-03
Division rather than addition for mm -> cm

dmbates:


knuesel:
Thanks, I guess that's my fault: in https://github.com/JuliaPlots/AlgebraOfGraphics.jl/issues/180 I used `+` instead of `/` to highlight the incompatibility with DataFrames.jl: `:bill_length_mm => (t -> t / 10) => "bill length (cm)"` works in DataFrames without broadcasting or `ByRow` so I changed it to `t + 10`, and it got copied over.

By the way on row 33 we could just write

```
df."bill length (cm)" = df.bill_length_mm / 10
```

instead of 

```
df.var"bill length (cm)" = map(t -> t / 10, df.bill_length_mm)
```

piever:
Well spotted, thanks for the PR!

> By the way on row 33 we could just write
> 
> ```julia
> df."bill length (cm)" = df.bill_length_mm / 10
> ```
> 
> instead of
> 
> ```julia
> df.var"bill length (cm)" = map(t -> t / 10, df.bill_length_mm)
> ```

I actually find `map` clearer even if it is unnecessary here, because that is how you would translate from AlgebraOfGraphics to DataFrames syntax in general.

knuesel:
> I actually find `map` clearer even if it is unnecessary here, because that is how you would translate from AlgebraOfGraphics to DataFrames syntax in general.

Makes sense (we could still get rid of the `var` prefix though).

PR #206, opened 2021-06-07
add FAQs section to docs

### LINKED ISSUE #180: Compatibility of pair syntax with DataFrames
It was noted on [Discourse](https://discourse.julialang.org/t/upcoming-algebraofgraphics-release/60968) that the pair syntax for mappings is close to the DataFrames but  note quite compatible. I thought it would be nice to have a place to summarize the issue and document the decision (hope you don't mind!).

**Single rows vs columns**

The main issue is that AlgebraOfGraphics mappings work on single rows instead of whole columns. For example this

```julia
:bill_length_mm => (t -> t + 10) => "bill length (cm)"
```
must be adapted for DataFrames to:

```julia
:bill_length_mm => (t -> t .+ 10) => "bill length (cm)"
```

or

```julia
:bill_length_mm => ByRow(t -> t + 10) => "bill length (cm)"
```

Advantages of single rows include:

* Simpler for the user in most cases.
* Less error prone especially
   * with grouped data (should a column operation apply to each group or the whole dataset?)
   * when several datasets are used

Downsides include
* No support for operations that require the whole column (for example z-score standardisation)
* Risk of confusion with DataFrames API
* Transformations cannot be reused between DataFrames and AlgebraOfGraphics

**Tuples vs arrays**

Another point I noticed in the machine learning tutorial section: AlgebraOfGraphics allows giving multiple input variables as a tuple:

```julia
error_rate = (:species, :predicted_species) => !isequal => "error rate"
```

In DataFrames only arrays are allowed:

```julia
error_rate = [:species, :predicted_species] => ByRow(!isequal) => "error rate"
```

I personally like arrays here as they stand out more in the middle of function calls, but that's subjective of course.

### LINKED ISSUE #184: Formalisation of algebra
I've been playing with the layer algebra a bit, and I'd be happy to try and write a formal description of the algebraic structure, which could serve as documentation. Would you be interested? It could also serve to check if the current structure has the desired properties.

What got me thinking about this is distributivity: in AlgebraOfGraphics `*` appears to distribute over `+`. But the algebra has the particularity that addition is non-commutative (the user can write `a+b` or `b+a` depending on which layer should be drawn on top). But distributivity generally implies the commutativity of `+` (see the simple proof [here](https://math.stackexchange.com/a/702626/549289) for example). So what's going on?

It turns out that `*` is right-distributive over `+`, but not left-distributive. Here's a counter-example, using layers that only contain positional mappings for simplicity:

```julia
# Concise print for layers, e.g. mapping(:a, :b) + mapping(:c) will print as (a,b) (c)
function Base.show(io::IO, s::AlgebraOfGraphics.OneOrMoreLayers)
    l::AlgebraOfGraphics.Layers = s
    print(io, join(["("*join(layer.positional, ",")*")" for layer in l.layers], " "))
end

# Define some simple layers
x = mapping(:x)
y = mapping(:y)
a = mapping(:a)
b = mapping(:b)
```

Distributivity works in simple cases:

```julia
julia> a*(x+y)
(a,x) (a,y)

julia> a*x + a*y
(a,x) (a,y)

julia> (x+y)*a
(x,a) (y,a)

julia> x*a + y*a
(x,a) (y,a)
```

But only right-distributivity works in more  complex cases:

```julia
julia> c = a+b
(a) (b)

julia> z = x+y
(x) (y)

julia> (x+y)*c
(x,a) (x,b) (y,a) (y,b)

julia> x*c + y*c           # Same as (x+y)*c
(x,a) (x,b) (y,a) (y,b)

julia> c*(x+y)
(a,x) (a,y) (b,x) (b,y)

julia> c*x + c*y           # Different from c*(x+y) !
(a,x) (b,x) (a,y) (b,y)
```

We could easily trade right-distributivity for left-distributivity by swapping the loops in https://github.com/JuliaPlots/AlgebraOfGraphics.jl/blob/dd8f9dc58f3e73b31c423e3e1bd5e7da311e1d5f/src/algebra/layers.jl#L22.

### LINKED ISSUE #205: `xscale` with densities looks strange
```julia
mwe = (; x = max.(0.1, randn(10000)))

plt = data(mwe) * mapping(:x => "x") * AlgebraOfGraphics.density()
draw(plt; axis = (; title="using `xscale=log10`", xscale=log10))
```
gives
 
77b7957d-ff88-4a02-a934-f353ebf0b413

while, using the same data,
```julia
plt = data(mwe) * mapping(:x => log10 => "log10(x)") * AlgebraOfGraphics.density()
draw(plt; axis = (; title="using transform in `mapping`"))
```
gives
915e4d45-5c35-46f9-ae87-9f31ff5a0c5f

The difference looks even more extreme in my non-MWE data. I think these should be the same, right? Is there a bug somewhere?

piever:
Fix #205. Fix #184. Fix #180.

### DIFF
--- a/docs/src/FAQs.md
+++ b/docs/src/FAQs.md
@@ -39,13 +39,13 @@ using AlgebraOfGraphics
 using AlgebraOfGraphics: density
 df = (x = exp.(randn(1000)),)
 kde1 = data(df) * mapping(:x) * density()
-draw(kde1, axis=(xscale=log,))
+draw(kde1, axis=(width=225, height=225, xscale=log,))
 ```
 
 ```@example logscaledensity
 df = (x = exp.(randn(1000)),)
 kde2 = data(df) * mapping(:x => log => "log(x)") * density()
-draw(kde2)
+draw(kde2, axis=(width=225, height=225))
 ```
 
 The two plots look different. The first represent the pdf of `x` in a log scale,
@@ -65,13 +65,13 @@ The same reasoning applies to histograms:
 using AlgebraOfGraphics
 df = (x = exp.(rand(1000)),)
 hist1 = data(df) * mapping(:x) * histogram()
-draw(hist1, axis=(xscale=log,))
+draw(hist1, axis=(width=225, height=225, xscale=log))
 ```
 
 ```@example logscalehist
 df = (x = exp.(rand(1000)),)
 hist2 = data(df) * mapping(:x => log => "log(x)") * histogram()
-draw(hist2)
+draw(hist2, axis=(width=225, height=225))
 ```
 
-Again, the data transformation approach is preferable as one obtains uniform bins.
+The data transformation approach is preferable as it produces uniform bins.
--- a/docs/src/FAQs.md
+++ b/docs/src/FAQs.md
@@ -2,19 +2,19 @@
 
 ## What is the algebraic structure of AlgebraOfGraphics?
 
-AlgebraOfGraphics is based on two operators, `+` and `*`. The algebraic structure these
-operators induce is that of a [semiring](https://en.wikipedia.org/wiki/Semiring), with a small
-caveat. Addition is commutative only up to the drawing order. For example `visual(Lines) + visual(Scatter)`
+AlgebraOfGraphics is based on two operators, `+` and `*`. These two operators induce
+a [semiring](https://en.wikipedia.org/wiki/Semiring) structure, with a small
+caveat. Addition is commutative only up to the drawing order. For example, `visual(Lines) + visual(Scatter)`
 is slightly different from `visual(Scatter) + visual(Lines)`, in that the former draws the scatter on top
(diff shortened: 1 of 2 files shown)


docs/src/FAQs.md
```diff
@@ -0,0 +1,80 @@
...
+proportionality relationship.
+
+A similar reasoning applies to histograms:
+
+```@example logscalehist
+using AlgebraOfGraphics
+df = (x = exp.(rand(1000)),)
+hist1 = data(df) * mapping(:x) * histogram()
+draw(hist1, axis=(width=225, height=225, xscale=log))
+```
```
ericphanson, line 71: just to say that I am pretty sure that https://github.com/JuliaPlots/Makie.jl/issues/1001 would fix this, i.e. choosing log-scale bins makes taking the `log` of the data the same as scaling the axis

PR #211, opened 2021-06-09
Update documentation for density

rikhuijzer:
As far as I understand, using `datalimits` is the only way to set the density range; I couldn't get it to work with `extrema`. 

This PR is just a temporary clarification. Maybe something with a `jldoctest` is better depending on how stable you expect this part of the code to be :)

### DIFF
--- a/src/transformations/density.jl
+++ b/src/transformations/density.jl
@@ -37,8 +37,9 @@ function (d::DensityAnalysis)(le::Entry)
 end
 
 """
-    density(; extrema, npoints, kernel, bandwidth)
+    density(; datalimits, npoints, kernel, bandwidth)
 
 Fit a kernel density estimation of `data`.
+For example, use `datalimits=((0.0, 60.0), )` to calculate the density from 0 to 60 on the x-axis.
 """
 density(; kwargs...) = Layer((DensityAnalysis(; kwargs...),))


src/transformations/density.jl
```diff
@@ -37,8 +37,9 @@ function (d::DensityAnalysis)(le::Entry)
 end
 
 """
-    density(; extrema, npoints, kernel, bandwidth)
+    density(; datalimits, npoints, kernel, bandwidth)
 
 Fit a kernel density estimation of `data`.
+For example, use `datalimits=((0.0, 60.0), )` to calculate the density from 0 to 60 on the x-axis.
```
piever, line 43:
```suggestion
For example, use `datalimits=((0.0, 60.0),)` to calculate the density from 0 to 60 on the x-axis.
```

While we're at it, maybe we could also mention that `npoints` is the number of points used by Makie to draw the line, whereas `kernel` and `bandwidth` are forwarded to `KernelDensity.kde`.

rikhuijzer:
This is how the docs look now:

image

I'm not sure what is the cause for the discontinued line.

Setting the ylabel to nothing is just a suggestion by me. I think that most people wouldn't understand `pdf` and even setting it to `density` seems a bit verbose because it's quite obvious that it's a density plot.

piever:
Thanks, PR looks great. You may have forgotten to push the `ylabel=""`, but I agree it makes sense, so that people can see that it is customizable and you don't have to stick with `"pdf"`.

It does look something is broken with the lines, do you get that weird behavior also locally? I'll look into it to try and fix it before merging this PR.

rikhuijzer:
> You may have forgotten to push the `ylabel=""`

What do you mean?

> It does look something is broken with the lines, do you get that weird behavior also locally?

It happens only locally! I've just generated the old (pre-PR) code locally and then I also get the weird behaviour. I didn't have any problems with similar plots anywhere yet, so for me this is not a problem.

I think we can merge. Let me know if there is anything to change still.

rikhuijzer:
> > You may have forgotten to push the `ylabel=""`
> 
> What do you mean?

Fixed.

Issue #224, opened 2021-07-01
Mixing categorical and continues colors doesn't work


SimonDanisch:
```julia
df = data([(a=1, b=0.1, c="a"), (a=2, b=0.2, c="b"), (a=3, b=0.3, c="c")])
draw(df * mapping(:a, color=:b) * visual(Scatter) + 
    df * mapping(:a, color=:c) * visual(Scatter))
```
Results in:
` Unknown color: a`

piever:
Closed by #226

PR #226, opened 2021-07-06
remove ContinuousScale

piever:


piever:
This actually turned out to be a larger refactor to ditch `ContinuousScale` altogether (we essentially do not use) and instead rely on native Makie for continuous scale support.

The main remaining item is `DateTime` support, which would need to be fixed on the Makie side.

piever:
Fixes #224

PR #228, opened 2021-07-07
color legend entry

SimonDanisch:
Legend currently doesn't pick up an e.g. markercolor:
```julia
df = (
    x = 1:10,
    y = rand(10),
    attr = ["m $i" for i in 1:10],
)
draw(data(df) * mapping(:x, :y, marker=:attr) * visual(Scatter, color=:red))
```
image

This PR takes uses the color, if it's a single color:
image

if multiple colors are used, it defaults back to black:
```julia
draw(data(df) * (mapping(:x, :y, marker=:attr) * (visual(Scatter, color=:red, markersize=20) + visual(Scatter, color=:green, markersize=10))))
```
image

### DIFF
--- a/src/guides/legend.jl
+++ b/src/guides/legend.jl
@@ -75,7 +75,6 @@ function _Legend_(fg::FigureGrid)
     for key in [:row, :col, :layout, :stack, :dodge, :group]
         pop!(scales, key, nothing)
     end
-
     # if no legend-worthy keyword remains return nothing
     isempty(scales) && return nothing
 
@@ -88,6 +87,11 @@ function _Legend_(fg::FigureGrid)
     labels = Vector{String}[]
     elements_list = Vector{Vector{LegendElement}}[]
 
+    # search in attributes if a color is given for e.g. lines/scatter
+    legend_colors = unique([entry.attributes[:color] for entry in entries(grid) if haskey(entry.attributes, :color)])
+    # if a single agreeable color is found, use that for the Legend entries
+    legend_color = length(legend_colors) == 1 ? first(legend_colors) : nothing
+
     for title in titles
         label_attrs = [key for (key, val) in scales if val.label == title]
         uniquevalues = mapreduce(k -> datavalues(scales[k]), assert_equal, label_attrs)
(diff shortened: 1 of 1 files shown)


greimel:
I don't fully understand what the code does. But what happens if use color for grouping but there happens to be just one group? (In my opinion, the legend for marker should be black despite there being just one color.)

What about a more general solution that adds all keywords from `visual` to `options` (which are then passed on to `legend_element`)?

(Not sure if that's easily doable, though)

src/guides/legend.jl
```diff
@@ -88,6 +87,11 @@ function _Legend_(fg::FigureGrid)
     labels = Vector{String}[]
     elements_list = Vector{Vector{LegendElement}}[]
 
+    # search in attributes if a color is given for e.g. lines/scatter
+    legend_colors = unique([entry.attributes[:color] for entry in entries(grid) if haskey(entry.attributes, :color)])
```
piever, line 91: I think this is not completely correct. What if one scatter layer has a color attribute, and another scatter layer has no color attribute, and is defaulting to the theme?

I think one should instead check that, for a given `plottype`, all the entries which use the `marker` scale (or whichever relevant scale is used in that legend group) also have a color attribute, and all the values agree. This may be not too hard to check, it should be all the entries where at least one of the `label_attrs`  is in the keys of `entry.primary`.

I'm also not sure if it should only be done for `color`, or there are other attributes in `visual` that should be considered.

piever:
> I don't fully understand what the code does. But what happens if you use color for grouping but there happens to be just one group? (In my opinion, the legend for marker should be black despite there being just one color.)

I think this PR only considers colors that are given via `visual`. I confess I'm in two minds about this, I'm not completely sure which is a better default for the legend, whether it is the theme default or the values it encounters in `visual` (should they all agree). Do you have some intuition as to which is more natural?

Somehow, both things seem reasonable to me, and maybe it's not too bad to briefly change the theme to have red markers (maybe `theme` can be passed as a keyword to `draw`, and is merged to the default theme).

jkrumbiegel:
this is fixed since 0.7

PR #233, opened 2021-07-13
Allow `draw!(axis, ...)`, and `mapping(markersize=:column_with_vec2)`

SimonDanisch:


src/algebra/layers.jl
```diff
@@ -74,14 +74,28 @@ function compute_axes_grid(fig, s::OneOrMoreLayers;
...
         return AxisEntries(ax, Entry[], scales)
     end
+    function create_axis(f::Axis, c)
+        if !isempty(axis)
+            @warn("Axis got passed, but also axis attributes. Ignoring axis attributes: $(axis)")
+        end
+        return AxisEntries(fig, Entry[], scales)
+    end
+    axs = compute_grid_positions(scales)
+    sizes = length.(axs)
```
piever, line 91:
```suggestion
    sizes = map(length, axs)
```

I try to consistently use `map` instead of `broadcast` for tuples (not sure if it makes a big difference though).

src/scales.jl
```diff
@@ -130,7 +130,7 @@ Determine whether `v` should be treated as a continuous or categorical vector.
 """
 function iscontinuous(u)
     v = Broadcast.broadcastable(u)
-    return !haszerodims(v) && eltype(v) <: Union{Number, Date, DateTime}
+    return !haszerodims(v) && eltype(v) <: Union{Number, Date, DateTime, Makie.StaticVector} && !(eltype(v) <: Bool)
```
piever, line 133: It may be better to set the `Makie.StaticVector` case as `isgeometry`, and then also exclude geometries in the `separate` function (when splitting categorical variables from the rest).

piever (changes requested):

piever:
Thanks for the PR and sorry for the long time to review! Left a few small comments but it mostly looks good!

Any idea why the docs are failing to build?

src/algebra/layers.jl
SimonDanisch: what difference do you expect?

piever: It's just an internal style guide to use `map` rather than `broadcast` for simple cases to avoid the compilation overhead of the broadcast machinery. That being said, definitely not a big deal, `broadcast` ends up calling `map` for tuples anyways.

src/algebra/layers.jl
```diff
@@ -74,14 +74,28 @@ function compute_axes_grid(fig, s::OneOrMoreLayers;
...
         type = get(axis, :type, Axis)
         options = Base.structdiff(axis, (; type))
         ax = type(fig[Tuple(c)...]; options...)
         return AxisEntries(ax, Entry[], scales)
     end
+    function create_axis(f::Axis, c)
+        if !isempty(axis)
+            @warn("Axis got passed, but also axis attributes. Ignoring axis attributes: $(axis)")
+        end
+        return AxisEntries(fig, Entry[], scales)
```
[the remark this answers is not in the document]
SimonDanisch: hmmm, how did that even work!? :D

piever (approved):

PR #238, opened 2021-07-20
Remove Axis(ae)

knuesel:
This removes the `Makie.Axis(ae::AxisEntries) = ae.axis` definition: it seems misleading to me as `Makie.Axis` is a constructor, and this is a getter.

I was confused when reading the example at http://juliaplots.org/AlgebraOfGraphics.jl/dev/generated/gallery/#Embedding-facets:  `Axis` was first used to create the axis on the left and that made sense, but I didn't understand why an axis was later created for each facet when rotating the label.

Using the `axis` field also seems consistent with the Makie API and the `figure` field advertised at http://juliaplots.org/AlgebraOfGraphics.jl/dev/FAQs/#How-to-combine-AlgebraOfGraphics-with-plain-Makie-plots.

### DIFF
--- a/docs/src/generated/gallery.jl
+++ b/docs/src/generated/gallery.jl
@@ -77,7 +77,7 @@ plt = data(df) * mapping(:x, :y, col=:i, row=:j)
 subfig = fig[1, 2:3]
 ag = draw!(subfig, plt)
 for ae in ag
-    Axis(ae).xticklabelrotation[] = π/2
+    ae.axis.xticklabelrotation[] = π/2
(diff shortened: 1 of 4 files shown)


piever:
Yes, I also think this is a good idea. It should probably go in the next breaking release where I was planning to stop overloading layoutable constructors (and use instead  `legend!` and `colorbar!`).

knuesel:
Sounds good! :)

piever:
Closing here as it is included in #249

PR #241, opened 2021-07-22
Use `RelocatableFolders` for font assets folders.

MichaelHatherly:
The current hardcoding of `@__DIR__` means that the package isn't usable with PackageCompiler when relocating system images since the source folders might not exist. This has come up before in Makie (where we solved it with `Artifacts`) and in `MathTeXEngine` (where we solved it with `RelocatableFolders` as done here). The `Artifacts` approach is a bit more involved to implement and maintain.


vjd:
@piever can we request a quick review

SimonDanisch:
@piever, I'll go ahead and merge + tag this, since the tests should be sufficient to "review" this ;) 
@MichaelHatherly, would it make sense to use the same for Makie? It looks like it does, and the artifact solution is pretty annoying ;)

MichaelHatherly:
> @MichaelHatherly, would it make sense to use the same for Makie? It looks like it does, and the artifact solution is pretty annoying ;)

Yes, pretty annoying I'd say! I'll sort it out.

piever:
Thanks for the PR and sorry for the delay in looking at this! 

LGTM, I'll merge and tag a patch release.

PR #254, opened 2021-08-10
respect drawing order

piever:


crstnbr:
Can this be merged? As I said on discourse already, I tested it locally and it worked nicely.

PR #258, opened 2021-08-11
set up gallery with DemoCards

johnnychen94:
This is a similar PR to https://github.com/JuliaPlots/PlotDocs.jl/pull/255 and I guess maybe you'll also be interested in it.

The demos are organized into folders, and the folder structure is mapped as the gallery sections/cards.

One caveat: The cover is generated on the fly, which means we have to run `demo.jl` once to generate the png file, and the second time to generate `demo.ipynb`, and finally Documenter will run it for the third time. Thus the build time might be 3x longer. I have plans to [make ipynb optional](https://github.com/johnnychen94/DemoCards.jl/issues/92) but this package is not of very high priority, to be honest.

There's a broken `docs/gallery/gallery/basic/makie_embedding.jl.bak` file, which I don't know how to generate the cover. Thus I'll leave it to you :)

The docs `make.jl` seems broken at present so I can't build it successfully, here's the result of `DemoCards.preview_demos("docs/gallery")`

Other live DemoCards examples:

- JuliaImages: https://juliaimages.org/dev/examples/
- Augmentor.jl: https://evizero.github.io/Augmentor.jl/dev/operations/


piever:
Thanks for the PR! I'm picking this up in #260 (so that I can do it together with some docs reorg I had in mind).

The build times did get a bit worse with this, it'd be really cool to at least save on the jupyter notebook generation. I guess making it optional is just a matter of adding a flag somewhere in DemoCards, or is it more complex than that?

The "broken docs" are because one of the democards (with geographic data) has a footnote, and in strict mode documenter throws an error (see https://github.com/JuliaPlots/AlgebraOfGraphics.jl/pull/258/checks?check_run_id=3302846071#step:6:29).

Can this be fixed on the DemoCards side somehow, to support cards with footnotes without throwing?

In non-strict mode, everything seems to work but with a warning. Unfortunately here the docs are also a "test suite", so I'd prefer to keep the strict mode.

PR #260, opened 2021-08-16
set up democards

piever:
Continuation of #258


docs/gallery/config.json
```diff
@@ -0,0 +1,3 @@
+{
+    "theme": "grid"
+}
```
johnnychen94, line 3: > The build times did get a bit worse with this, it'd be really cool to at least save on the jupyter notebook generation. I guess making it optional is just a matter of adding a flag somewhere in DemoCards, or is it more complex than that?

This doesn't work right now, but would disable all notebook generation in future DemoCards release https://github.com/johnnychen94/DemoCards.jl/pull/98

```suggestion
{
    "theme": "grid",
    "properties": {
        "notebook": "false"
    }
}
```

johnnychen94 (commented):
> Can this be fixed on the DemoCards side somehow, to support cards with footnotes without throwing?

The Makie example is skipped as I don't know how to generate the cover for it. As long as it's fixed and renamed to `.jl` from `.jl.bak`, I believe it should be okay.

docs/gallery/config.json
piever: Great! I'll wait for the next DemoCards release to add this

johnnychen94:
Just a small trick: I use `preview_demos("path_to_demo_file.jl")` or `preview_demos("path_to_demo_folder/")` a lot when I'm trying to add or modify only one or few files. More in https://johnnychen94.github.io/DemoCards.jl/dev/preview/

docs/gallery/config.json
johnnychen94: In case I didn't make it clear, it's harmless to commit this change as it won't take any effect or break the current workflow right now.

johnnychen94: Well. Let me make a release for it

piever: Ah, of course, I can just merge this patch here and it will get faster as soon as your release goes through. Thanks a lot for the feature and the release!

PR #261, opened 2021-08-16
fix integer hist and add to gallery

piever:


storopoli:
This is great I have a lot of scripts that do a `float` transformation before AoG `histogram`. No more!

PR #268, opened 2021-08-20
show how to modify a AoG plot using Makie

tbenst:
Thought it'd be useful to highlight that AoG plots are "just" Makie

### DIFF
--- a/docs/src/FAQs.md
+++ b/docs/src/FAQs.md
@@ -119,3 +119,21 @@ fg #hide
     When setting the `width` and `height` dimensions of each axis manually you
     will need to call `AlgebraOfGraphics.resizetocontent!(fg)` before
     displaying the figure such that each axis is sized correctly.
+
+This `.figure` field can also be used to adjust a plot using `Makie`.
+See the Makie [API reference](http://makie.juliaplots.org/stable/makielayout/reference.html)
+for more information.
+
+```@example combined-3
+using AlgebraOfGraphics, CairoMakie #hide
+
+df = (x = exp.(rand(1000)),) #hide
+hist2 = data(df) * mapping(:x => log => "log(x)") * histogram()  #hide
+fg = draw(hist2; figure = (resolution = (600, 400),))  #hide
+
+lines(fg.figure[1, 1], 0..2pi, cos) #hide
+fg.figure.content[1].xticklabelrotation = Float64(pi/6)
+fg #hide
+```
+
+See the 


docs/src/FAQs.md
```diff
@@ -119,3 +119,21 @@ fg #hide
...
+
+```@example combined-3
+using AlgebraOfGraphics, CairoMakie #hide
+
+df = (x = exp.(rand(1000)),) #hide
+hist2 = data(df) * mapping(:x => log => "log(x)") * histogram()  #hide
+fg = draw(hist2; figure = (resolution = (600, 400),))  #hide
+
+lines(fg.figure[1, 1], 0..2pi, cos) #hide
+fg.figure.content[1].xticklabelrotation = Float64(pi/6)
```
piever, line 135: Thanks for the example! I was thinking that it would be better IMO if here one was modifying the figure directly rather than the axes (the axes should ideally be accessed as `fg.grid[i, j].axis`). Maybe one could change `backgroundcolor` for instance, or any other figure attribute?

docs/src/FAQs.md
```diff
@@ -119,3 +119,21 @@ fg #hide
...
+for more information.
+
+```@example combined-3
+using AlgebraOfGraphics, CairoMakie #hide
+
+df = (x = exp.(rand(1000)),) #hide
+hist2 = data(df) * mapping(:x => log => "log(x)") * histogram()  #hide
+fg = draw(hist2; figure = (resolution = (600, 400),))  #hide
+
+lines(fg.figure[1, 1], 0..2pi, cos) #hide
```
piever, line 134: To build on top of the previous example, you should be able to just do `@example combined-2`, without having to copy and hide what comes before.

tbenst:
apologies--I will get to this, crunching for a paper deadline...

tbenst:
adding a note-to-self to also contribute something on how to change ordering since took me a while to figure out
```julia
using PalmerPenguins, DataFrames
penguins = dropmissing(DataFrame(PalmerPenguins.load()))
penguins.island = CategoricalArray(penguins.island, ordered=true)
levels!(penguins.island, ["Dream", "Biscoe", "Torgersen"])
first(penguins, 6)
penguin_frequency = aog.data(penguins) * frequency() * mapping(:species)
plt = penguin_frequency * mapping(color = :island, dodge = :island)
draw(plt)
```

image

docs/src/FAQs.md
asinghvi17:
```suggestion
only(contents(fg.figure[1, 1])).xticklabelrotation = Float64(pi/6)
```
is what the current Makie API prescribes I think.

jkrumbiegel:
there's now a tutorial that shows these aspects https://aog.makie.org/v0.10.0/tutorials/intro-vi

PR #278, opened 2021-09-15
Simple documentation update to use capital letters for first argument

flyaflya:


docs/src/generated/visual.jl
```diff
@@ -3,7 +3,8 @@
 # The function `visual` can be used to give data-independent visual information about the plot
 # (plotting function or attributes).
 
-# The available plotting functions are documetned [here](http://makie.juliaplots.org/dev/examples/plotting_functions/).
+# The available plotting functions are documented [here](http://makie.juliaplots.org/dev/examples/plotting_functions/).
+# Capitalize the first-letter of the function to use it as an argument to `visual` (e.g. `visual(Scatter)`).
```
piever, line 7: Maybe it would be best to mention that it's actually CamelCase, so for example `barplot` becomes `BarPlot`.

piever:
Sorry for the slow reply! I've left a small comment.

flyaflya (commented):
made slight mod as requested.  thx!

docs/src/generated/visual.jl
flyaflya: good point :-)

piever:
Great, thank you

PR #289, opened 2021-10-11
Support LaTeX axis labels

greimel:
replaces #284.

Defer the creation of the `Axis` until after the axis labels are known.

Introduce an intermediate struct `_AxisGrid_` that doesn't store an `Axis`, but the axis type, figure position and options.

image

see also here: http://juliaplots.org/AlgebraOfGraphics.jl/previews/PR289/gallery/gallery/data%20manipulations/new_columns_on_the_fly/#Renaming-(including-LaTeX-labels)-and-transforming-variables


piever:
Closing here as it will be merged as a part of #316

PR #290, opened 2021-10-11
Move colorbar stuff to its own file

greimel:
to make space for improved legend functionality


piever:
Good idea, given that the legend code will become a fair bit more complicated. Merging here so that #291 is easier to review.

Issue #321, opened 2022-01-06
Docstrings for objects defined with `@recipe` throw an error


haberdashPI:
As a new user of AlgebraOfGraphics, an obvious thing to do if you want to understand more of the details of `visual` is by calling e.g. `?Lines` at the REPL to know more about how to use specific visualizations. It is not always obvious how the APIs between these objects map to those that are referenced in the actual docs (the original Makie functions each recipe is based on)

But, this is what happens when you try to do that:

```julia
help?> Lines
search: Lines lines lines! LinesFill linesfill linesfill! LineSegments linesegments linesegments!

ERROR: UndefRefError: access to undefined reference
Stacktrace:
 [1] getproperty(x::Type, f::Symbol)
   @ Base ./Base.jl:37
 [2] summarize(io::IOBuffer, TT::Type, binding::Base.Docs.Binding)
   @ REPL /usr/local/julia/share/julia/stdlib/v1.7/REPL/src/docview.jl:286
 [3] summarize(binding::Base.Docs.Binding, sig::Type)
   @ REPL /usr/local/julia/share/julia/stdlib/v1.7/REPL/src/docview.jl:249
 [4] doc(binding::Base.Docs.Binding, sig::Type)
   @ REPL /usr/local/julia/share/julia/stdlib/v1.7/REPL/src/docview.jl:182
 [5] doc(binding::Base.Docs.Binding)
   @ REPL /usr/local/julia/share/julia/stdlib/v1.7/REPL/src/docview.jl:160
 [6] macro expansion
   @ /usr/local/julia/share/julia/stdlib/v1.7/REPL/src/docview.jl:470 [inlined]
 [7] top-level scope
   @ ~/.julia/packages/Infiltrator/4K3Nh/src/Infiltrator.jl:484
 [8] top-level scope
   @ ~/.julia/packages/CUDA/YpW0k/src/initialization.jl:52
```

Seems like that should be fixable. Would be happy to try and dig into it a little myself if you don't have a plan for that already.

piever:
I confess I don't have a lot of bandwidth to look into this at the moment, it would be great if you could look into this. I suspect it is just an issue of the `@recipe` macro to be fixed in Makie directly.

haberdashPI:
Ah, I didn't that was something from Makie, I'll take ask over there.

piever:
Closing here in favor of https://github.com/JuliaPlots/Makie.jl/issues/1589

PR #324, opened 2022-01-14
Clarify docs for `visual`

haberdashPI:
I've added some clarifications for how keyword arguments are passed (it was not obvious to me until I read the source)

I've indicated that any `@recipe` created plot function can be used with `visual`, since this was also not obvious to me until I read the source / read the response in #321 

By the way, thanks for creating this awesome package!


piever:
Thanks for the nice words and the PR! Yes, the clarification is very useful, because it is one of the design goals that custom recipes work out of the box with AlgebraOfGraphics.

piever (approved):

Issue #329, opened 2022-01-26
Colormap doesn't work on lines


haberdashPI:
At the very least, the obvious thing you would want to do (set `colormap` in visual) doesn't seem to work. 

```julia
x = rand(100, 4)
df = DataFrame(t = repeat(axes(x, 1), outer=4), x = vec(x), kind=repeat(["a", "b", "c", "d"], inner=100))
plt = data(df) * mapping(:t, :x, color = :kind) * visual(Lines, colormap = :Paired_4)
save("test.png", draw(plt))
```

This yield output that is identical to the output when you remove `colormap`.

In contrast, the following Makie-only version appropriately colors the lines.

```julia
x = rand(100, 4)
df = DataFrame(t = repeat(axes(x, 1), outer=4), x = vec(x), kind=repeat(["a", "b", "c", "d"], inner=100))
fig = Figure()
Axis(fig[1,1])
for sdf in groupby(df, :kind)
    lines!(sdf.t, sdf.x, color=sdf.kind, colorrange = (1, 4), colormap=:Paired_4)
end
save("test.png", fig)
```

Here's the output of `]st`

```
  [cbdf2221] AlgebraOfGraphics v0.6.0
  [13f3f980] CairoMakie v0.6.6
  [a93c6f00] DataFrames v1.3.2
  [ee78f7c6] Makie v0.15.3
```

palday:
Same seems to hold for `Scatter`.

danielw2904:
I think I am hitting a similar issue. Here the `colormap` is completely ignored:

```
dd = DataFrame(x = [rand(10)..., 10 .+ rand(10)...], c = [repeat('a', 10)..., repeat('b', 10)...])
plt = data(dd) * 
mapping(:x, color = :c) * 
AlgebraOfGraphics.density()
fig = draw(plt * visual(colormap = :reds))
```

piever:
I think the issue is that categorical color variables (eg, columns of strings) are converted using the default color palette, so AlgebraOfGraphics does actually not look into `colormap` at all.

For what you are looking for, you should probably do

```julia
draw(plt, palettes=(color=cgrad(:Paired_4),))
```

(see [last example here](http://juliaplots.org/AlgebraOfGraphics.jl/stable/gallery/gallery/scales/custom_scales/#Custom-scales)).

[...]

PR #344, opened 2022-02-08
document apply_palette

piever:
Marked as draft as we may go for a different approach, described in #329


jkrumbiegel:
This has changed in 0.7 so I'll close this

PR #375, opened 2022-03-22
Add example of colormap tweaking

KronosTheLate:
An example I had to go digging deeper for.


piever:
Thanks! Merged it as part of #377

PR #378, opened 2022-04-15
Bump Makie and GridLayoutBase version

rongcuid:
The new version of Makie fixes https://github.com/JuliaPlots/Makie.jl/issues/1748. `Layoutable` has been renamed to `Block`.


piever:
Closing here (bumped GridLayoutBase in #379 instead). In general, package compatibility should be given with `=` (instead of `>=` or no compat at all) to be consistent with semantic versioning. So for example, when Makie 0.17 comes out, with the breaking `Layoutable => Block` rename, one would do the rename here and change compat to `Makie = 0.17`.

PR #383, opened 2022-04-25
Update compat bounds for Makie

haberdashPI:
When testing out some change to make [RainClouds](https://github.com/JuliaPlots/Makie.jl/pull/1725) usable with AoG, I needed to update some compat bounds. I ran tests, and everything seems to be working okay.


greimel:
This shouldn't be necessary. Doesn't the compat entry `0.x.y` include all `0.x.z` for `z >= y`?

haberdashPI:
So it does...

PR #384, opened 2022-04-26
Support incomplete list for `sorter` and `renamer`

haberdashPI:
This addresses the proposal in #329 to make it possible to handle palettes a little more cleanly:

Specifically the missing piece this adds is to handle the case where sorter and renamer are provided only a partial enumeration of values.

Then, to quote @piever, the old:

```julia
# `a` and `b` entries get specified colors, for the other entries it's data-dependent
palette = (color=["a" => "red", "b" => "blue", "green", "black"],)
```

can now be achieved with

```julia
mapping(color=v => sorter(["a", "b"])
palette = (color=["red", "blue", "green", "black"],)
```

### DIFF
--- a/docs/gallery/gallery/scales/custom_scales.jl
+++ b/docs/gallery/gallery/scales/custom_scales.jl
@@ -42,15 +42,15 @@ plt = data(df) *
     visual(Arrows, arrowsize=10, lengthscale=0.3)
 draw(plt; palettes=(arrowcolor=colors, arrowhead=heads))
 
-# To associate specific attribute values to specific data values, use pairs.
-# Missing keys will cycle over values that are not pairs.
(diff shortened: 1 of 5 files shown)


piever:
About the implementation, instead of overloading `map`, I was thinking that one could try the following. `renamer` and `sorter`, rather than returning a `Sorted` object, would return a
```julia
struct Renamed{S, T}
    original::S
    value::T # maybe `Union{T, Nothing}` is cleaner, otherwise just a random one if the correct one is not found
    idx::UInt32 # use `0` if `original` is not found in `uniquevalues`
end
```
object. Then, this gets all simplified to something reasonable using `rescale` at the very end.

haberdashPI:
Ah yes, I see, that is probably a better way to do it!

Above is just an initial commit/sketch. It's untested, and I haven't added any new tests yet.

PR #395, opened 2022-05-26
Widen `GLM` compat bounds.

MichaelHatherly:
I'm sure CompatHelper would get around to opening one for this quite soon, however I was in need of using some specific versions together. Tests all pass locally. 1.7.1.8 GLM changes don't look particularly significant in relation to what would affect AoG.


piever:
Are you sure it is not automatically supported already? I thought `"1.7"` meant that we accept all `1.x.y` versions no older than `1.7.0`.

MichaelHatherly:
Whoops, it was something else holding it back, `"1.7"` does indeed allow for `1.8` as well of course.

PR #408, opened 2022-06-27
Pagination

jkrumbiegel:
This PR adds pagination as discussed in https://github.com/JuliaPlots/AlgebraOfGraphics.jl/issues/303.

The approach is pretty much what @MichaelHatherly detailed in that post, I made some smaller changes and added tests and docs. Layer objects are split into groups depending on the maximum number of rows/columns/wrapped facets allowed and therefore `draw` returns a `Vector{FigureGrid}`. There's also a `draw(paginated, i)` method in case not all figures should live in memory at the same time, this method returns a `FigureGrid` accordingly.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,6 @@ PooledArrays = "2dfb63ee-cc39-5dd5-95bd-886bf059d720"
 RelocatableFolders = "05181044-ff0b-4ac5-8273-598c1e38db00"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
-TableOperations = "ab02a1b2-a7df-11e8-156e-fb1833f50b87"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
--- a/Project.toml
+++ b/Project.toml
@@ -20,6 +20,7 @@ PooledArrays = "2dfb63ee-cc39-5dd5-95bd-886bf059d720"
 RelocatableFolders = "05181044-ff0b-4ac5-8273-598c1e38db00"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
+TableOperations = "ab02a1b2-a7df-11e8-156e-fb1833f50b87"
 Tables = "bd369af6-aec1-5ad0-b16a-f7cc5008161c"
 
 [compat]
--- a/src/algebra/layers.jl
+++ b/src/algebra/layers.jl
@@ -78,7 +78,7 @@ function compute_palettes(palettes)
     return foldl(merge!, (layout, theme_palettes, user_palettes), init=NamedArguments())
 end
 
-function compute_axes_grid(fig, s::OneOrMoreLayers;
+function compute_axes_grid(fig, s::Union{OneOrMoreLayers, ProcessedLayers};
                            axis=NamedTuple(), palettes=NamedTuple())
 
     axes_grid = compute_axes_grid(s; axis, palettes)
@@ -92,12 +92,18 @@ function compute_axes_grid(fig, s::OneOrMoreLayers;
 end
 
 function compute_axes_grid(s::OneOrMoreLayers;
-                           axis=NamedTuple(), palettes=NamedTuple())
-    layers::Layers = s
-    processedlayers = map(ProcessedLayer, layers)
+    axis=NamedTuple(), palettes=NamedTuple())
 
+    layers::Layers = s
+    processedlayers = ProcessedLayers(map(ProcessedLayer, layers))
+    compute_axes_grid(processedlayers; axis, palettes)
+end
+
+function compute_axes_grid(p::ProcessedLayers;
+                           axis=NamedTuple(), palettes=NamedTuple())
     palettes = compute_palettes(palettes)
 
+    processedlayers = p.layers
     categoricalscales = MixedArguments()
     for processedlayer in processedlayers
         mergewith!(
--- a/src/algebra/layers.jl
+++ b/src/algebra/layers.jl
@@ -92,7 +92,7 @@ function compute_axes_grid(fig, s::Union{OneOrMoreLayers, ProcessedLayers};
 end
 
 function compute_axes_grid(s::OneOrMoreLayers;
-    axis=NamedTuple(), palettes=NamedTuple())
+                           axis=NamedTuple(), palettes=NamedTuple())
 
     layers::Layers = s
     processedlayers = ProcessedLayers(map(ProcessedLayer, layers))
--- a/src/paginate.jl
+++ b/src/paginate.jl
@@ -41,7 +41,6 @@ function paginator(layer::Layer, data, spec, limiter)
     layers = map(sets) do set
         return Layer(layer.transformation, func(set)..., layer.named)
     end
-    Main.@infiltrate
     return Paginate(layers)
 end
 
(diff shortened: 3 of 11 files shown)


jkrumbiegel:
I didn't know about ProcessedLayer, but it sounds useful to avoid handling differences in input arguments this way. That analyses are already computed means that the `draw(..., i)` command is a bit less useful as more work is being done already, but it's probably still good to keep it for situations where you don't want to create all Makie plots at the same time.

So I can build ProcessedLayers and pass those to draw just like with Layers otherwise?

> On an unrelated note

Of course, I'll adjust the format

piever:
> So I can build ProcessedLayers and pass those to draw just like with Layers otherwise?

This may require a tiny refactor. There is a `compute_axes_grid` [method](https://github.com/JuliaPlots/AlgebraOfGraphics.jl/blob/cbe471dba5e37f420269b6a470b851bee91fd2e7/src/algebra/layers.jl#L94) where the conversion to `ProcessedLayer` happens (and it is essentially where the nontrivial part of `draw` lives). So it should be a matter of allowing `compute_axes_grid` to receive one or more `ProcessedLayer`s, and widening the signature of `draw` & co. a little bit.

If you need more fine-grained control over this, `grid = compute_axes_grid(fig, layers; kws...)` returns a grid of objects with all the information (including the reference to the figure position) to plot things. Calling `foreach(plot!, grid)` will actually draw the plots (see eg [here](https://github.com/JuliaPlots/AlgebraOfGraphics.jl/blob/cbe471dba5e37f420269b6a470b851bee91fd2e7/src/draw.jl#L16)).

jkrumbiegel:
Ok but is that part of the pipeline already connected to drawing? I thought it made most sense to split on the data level, as if each pagination was just the same plot with less data. So I don't want to interact with Makie for that.

jkrumbiegel:
I've rewritten the code to use ProcessedLayers. I'm not sure if I've understood the grouping aspect correctly, the code seems to work but I still had to subset manually, while it sounded like you didn't think that was necessary. Maybe there's another way?

> different pages of the pagination might have e.g. different colorranges or maybe even different legends

That's true. @MichaelHatherly what do you think about that aspect? Axis limits, color legend categories, etc, can all differ across pages.

MichaelHatherly:
> Axis limits, color legend categories, etc, can all differ across pages.

If I recall correctly, `ggforce`, where the paginate idea came from for this, does maintain consistency of things like axis limits across each resulting page of plots. Though it might not matter too much in practice if it does or not for our `paginate`, probably worth us trying out some complex data and seeing how it looks.

piever:
> I'm not sure if I've understood the grouping aspect correctly, the code seems to work but I still had to subset manually, while it sounded like you didn't think that was necessary.

Ah, what I meant is that the data is already grouped at that stage, which in my mind would simplify things, but it's true that subsetting is still somewhat manual, you are right.

I'm planning a minor refactor of the code from `ProcessedLayers` to the grid of subplots. I suspect that ideally you would intercept things somewhere in the middle of `compute_axes_grid`: before you create the grid (as that depends on the pagination), but after you have fitted categorical and continuous scales.

Maybe a pragmatic approach is to merge something similar to the current implementation, and keep in mind when refactoring that bit that we want to be able to simplify the pagination code and fix the issue with scale consistency. (Might be best to add a `# FIXME` in the code.)

Another (possibly less practical) alternative would be to work with the output of [this method](https://github.com/JuliaPlots/AlgebraOfGraphics.jl/blob/b5ff5f648b0b1ed0036602da49eb2130fc92b96d/src/algebra/layers.jl#L94), which basically gives you a grid of axis specifications and plot specifications that abstractly represents the full facet plot (Makie hasn't been involved yet). There you get consistent legends and colorbars, but the annoying bit is that the facet grid has been computed already, so you would have to manually recompute it based on pagination, which seems not ideal.

jkrumbiegel:
Hi Pietro, checking in again to see what would need to be done to get this or some alternative finished.

As far as I understand, the main drawback of the current implementation is that scales are not synchronized between pages because each page works with a separate subset of the data, right? Also, this means that layouts are not synchronized either (not sure if that's desirable anyway) because if the last page has fewer Axes they will be bigger.

But these things don't really affect the API proposed here, so we could for example introduce a different implementation with a default kwarg `synchronize_pages = false` which could be set to `true`. In that case the update would not be breaking.

Project.toml
```diff
@@ -20,6 +20,7 @@ PooledArrays = "2dfb63ee-cc39-5dd5-95bd-886bf059d720"
 RelocatableFolders = "05181044-ff0b-4ac5-8273-598c1e38db00"
 StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
+TableOperations = "ab02a1b2-a7df-11e8-156e-fb1833f50b87"
```
piever, line 23: Maybe this is not needed anymore?

src/paginate.jl
```diff
@@ -0,0 +1,185 @@
...
+        col = nothing,
+    )
+    layers = l isa Layer ? Layers([l]) : l
+    processed_layers = map(ProcessedLayer, layers.layers)
+    inverted = [paginate_layer(processed_layer; layout, row, col) for processed_layer in processed_layers]
+    layers = map(ProcessedLayers, invert(inverted))
+    return PaginatedLayers(layers, layout, row, col)
+end
+
+# copied from SplitApplyCombine.jl to avoid full dependency
```
piever, line 155: I understand that the fully optimized implementation is a bit complex, but couldn't we just do

```julia
function invert(a::AbstractArray{<:AbstractArray})
    f = first(a)
    all(axes(v) == axes(f) for v in a) || error("keys don't match")
    map(i -> [v[i] for v in a], keys(f))
end
```

Even though I'm sure it's been tested thoroughly in SplitApplyCombine.jl, I'd rather keep a simple implementation here.

piever (changes requested):

piever:
Hey Julius! Yes, if we are OK with the scales not being shared (which can be added later), then this can probably be merged as is (I've left a couple small comments). The scales issue is probably best addressed in a future refactor of the data structures holding the relevant information (at the moment it seemed very inconvenient to do).

Re: changing the API in the future, I wonder whether it may be best to add a note in the `paginate` docs that the exact details of scales / layout synchronization are not set in stone and may change with non-breaking releases.

src/paginate.jl
jkrumbiegel: Agreed

jkrumbiegel:
Ok I have incorporated your suggestions. On my machine the pagination tests succeed but some other tests fail (I think because of a backend change in Axis linking a while ago in Makie).

I think it's good to start with this version as experimental and then add fancier behavior that requires refactoring later. The plots on each page are not "wrong" if they're not synchronized across pages, it might just be something that people want.

piever (approved):

piever:
> On my machine the pagination tests succeed but some other tests fail (I think because of a backend change in Axis linking a while ago in Makie).

Ah yes, those were fixed on the master branch already.

jkrumbiegel:
Great, I assume you'll merge this when convenient, I have rights to do it but don't feel like that's what I should do :)

src/algebra/layers.jl
```diff
@@ -92,12 +92,18 @@ function compute_axes_grid(fig, s::OneOrMoreLayers;
 end
 
 function compute_axes_grid(s::OneOrMoreLayers;
-                           axis=NamedTuple(), palettes=NamedTuple())
+    axis=NamedTuple(), palettes=NamedTuple())
```
piever, line 95:
```suggestion
                           axis=NamedTuple(), palettes=NamedTuple())
```

piever:
> Great, I assume you'll merge this when convenient, I have rights to do it but don't feel like that's what I should do :)

Yes, I am merging today so that I have time to also tag a release with the feature

PR #417, opened 2022-07-25
ignore order in axes linking tests

daschw:
Sets are now used in Makie when linking axes (https://github.com/JuliaPlots/Makie.jl/pull/2116/files#diff-0e352caa51e7f7cbfbdfc6e56c0ded4d3ed714cf9770d011c283cbb561eb3e49R1014) so order is not preserved anymore for linked axis. Hence, the AOG tests should not compare vectors, but sets. This fixes the failing tests.


piever:
Thank you!

piever (approved):

PR #432, opened 2022-10-25
Export Layers type

nathanrboyer:
`Layer` and `ProcessedLayer` are already exported. Why not `Layers`?

I need this in order to store plots in a custom struct:
```julia
struct MyType
    p1::Layer
    p2::Layers
end
```


piever:
Makes sense, not exporting it was most likely an oversight.

piever (approved):

PR #438, opened 2022-11-29
Use SnoopPrecompile

### LINKED ISSUE #430: SnoopPrecompile
Came up in [this discourse thread](https://discourse.julialang.org/t/very-slow-time-to-first-plot-2022/88968/48). AOG doesn't have snoop-precompile statements (i.e. a typical workflow script as part of precompilation) which is why generating a system image containing it doesn't give the desired (and possible) speedups. Please consider using SnoopPrecompile! It might also improve TTFX without system images 😉 

(See e.g. OrdinaryDiffEq for an example: https://github.com/SciML/OrdinaryDiffEq.jl/blob/v6.27.1/src/OrdinaryDiffEq.jl#L207-L333)

greimel:
closes #430 

```julia
begin
       df(N) = (x = rand(N), y = rand(N))
       t = time()
       using AlgebraOfGraphics, CairoMakie
       @info time() - t
       fig = data(df(100)) * mapping(:x, :y) * visual(Scatter) |> draw
       @info time() - t
       save("/tmp/aog.pdf", fig)
       @info time() - t
end
```

## Timings on https://github.com/JuliaLang/julia/pull/47184

* master:
```
[ Info: 9.63295292854309
[ Info: 26.53355598449707
[ Info: 28.348171949386597
```
* PR: 
```
[ Info: 11.315019130706787
[ Info: 19.168047189712524
[ Info: 21.042071104049683
```

* Note that `using` time goes up, but total time goes down.
* The timings a probably unfair because there is a `Scatter` in the precompiles
* I did a few timings, and they were always within a second of the shown times.


piever (approved):

piever:
Thanks for this, it's very useful. I imagine in the future one might wish to tweak what exactly gets precompiled, but that seems like a good start.

PR #441, opened 2022-12-04
documentation: tutorial, penguins example

cecileane:
I just started learning about `AlgebraOfGraphics`, in the hope of switching over from `ggplot2` in `R`. I love it! I could use more help from the documentation. These [tutorials](https://tutorials.pumas.ai/html/PlottingInJulia/) have helped a lot for things I could not find here.
This PR is to fix 2 "errors" I ran across as a beginner, when learning to use AoG.

Here is a small example to reproduce the first error:

```julia
using AlgebraOfGraphics, CairoMakie
set_aog_theme!()
df1 = (x=[1,2,3,1,2,3], y=[4,3,1,5,3,2])
plt = data(df1) * mapping(:x,:y)

# replicating the tutorial example works:
axis = (width = 225, height = 225)
draw(plt; axis)

# but this fails, if I rename the variable 'axis'
# which was counterintuitive (at least for me not being familiar with Makie / AoG)
ax = (width = 225, height = 225)
draw(plt; ax)
# ERROR: MethodError: no method matching draw(::Layer; ax=(width = 225, height = 225)) ...
```
The suggested change makes it more explicit that `axis` is a keyword option, and helps to learn how to tweak figures.

The second "error" is that the png figure created with the `save(...)` example contains a tiny part of the plot only (top left corner). The suggested change doesn't have this issue, but uses svg instead. I don't know how to fix the issue with png, sorry (still learning!).

```julia
fg = draw(plt; axis=(; width = 225, height = 225))
save("figure.png", fg, px_per_unit = 3) # to reproduce the problematic (too small) png file
save("figure.svg", fg) # the svg file looks good.
```

### DIFF
--- a/docs/src/generated/penguins.jl
+++ b/docs/src/generated/penguins.jl
@@ -43,8 +43,8 @@ draw(penguin_frequency; axis=(; width = 225, height = 225))
 # inspect it in the file explorer.
 #
 # ```julia
-# fg = draw(penguin_frequency; axis)
-# save("figure.png", fg, px_per_unit = 3) # save high-resolution png
+# fg = draw(penguin_frequency; axis=(; width = 225, height = 225))
+# save("figure.svg", fg)
 # ```
 #
 # ### Styling by categorical variables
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Thank you! :) Great to have the docs more user friendly!

The problem with png output is an actual bug, which should be fixed by: 
https://github.com/MakieOrg/Makie.jl/pull/2470

piever:
Thank you for the PR!

I see that the special syntax `draw(penguin_frequency; axis)` could be confusing if one does not know the rule, but I think we should go for a consistent alternative.

I would suggest one of the following options
1. replace the variable declaration `axis = (width = 225, height = 225)` with `ax = (width = 225, height = 225)` and update all the calls to be `draw(...; axis=ax)`, or
2. leave the variable declaration `axis = (width = 225, height = 225)` as is, but use the explicit form `draw(...; axis=axis)`
3. leave the code as is, but add a small note at the beginning about the special syntax `draw(...; axis)`.

Re pngs: ideally that part can be reverted now that the upstream bug is fixed?

cecileane:
That makes total sense. I went for option 2, which seemed best for a user with little Julia experience (compared to option 3). Also reverted the figure format to png.

piever (approved):

piever:
Perfect, thanks!

PR #443, opened 2023-01-06
fix deprecation warning

jeremiahpslewis:


src/AlgebraOfGraphics.jl
```diff
@@ -9,13 +9,18 @@ using GeoInterface: coordinates, isgeometry, geomtrait, PolygonTrait, MultiPolyg
...
-                          linkaxes!,
-                          linkxaxes!,
-                          linkyaxes!
+using Makie: current_default_theme,
+             to_value,
+             automatic,
+             Automatic,
+             PlotFunc,
+             ATTRIBUTES,
+             hidexdecorations!,
```
piever, line 18: These methods are being extended, so one should continue importing them.

jeremiahpslewis:
@piever I think this is good to go. :)

SimonDanisch (approved):

PR #447, opened 2023-03-17
Linear extension for polynomial regression

FedeClaudi:
Hey everyone,

Thank you for the awesome work with this project.
Here I propose a small change to `LinearAnalysis` to include `x^i` terms and allow for polynomial regression. 
I currently use this as a separate layer in my own code, but I thought it would be a useful addition to AOG.

Thank you, 
Federico


jkrumbiegel:
Thanks for the PR, just getting to take a look at it now. Could you maybe demonstrate visually what this is supposed to do? I'm not sure what the benefit of `x^i` terms would be in this context. And we'd need to add an illustrative example to the docs anyway.

PR #461, opened 2023-05-16
Show styles from mapping in legends from `ScatterLines`, `VLines`, `HLines` and `ABLines`

### LINKED ISSUE #435: Legend doesn't properly show linestyle
Instead of showing the line styles, it just shows gray blobs:

```julia
using AlgebraOfGraphics # v0.6.11
using CairoMakie # v0.8.13

x = 1:3
y = 'a':'c'
df = (; x, y)
plt = data(df) * mapping(:x, linestyle=:y) * visual(VLines)

fg = draw(plt)
save("mwe.png", fg)
```

mwe

The behavior is the same for AlgebraOfGraphics `v0.6.12` and CairoMakie `v0.9.2`.

jotas6:
Hello! I recently tried to generate a plot with `ScatterLines` specifying `mapping(...; marker=:var, linestyle=:var)`, but the resulting legend didn't show the styling appropriately. I noticed that this was pointed out on #435, along with a similar issue for `VLines`, so I followed @skleinbo's recommendation and modified `src/guides/legendelements.jl` the following way:

- Changed `legend_elements(::Type{Lines}; ...)` to `legend_elements(::Union{Type{Lines}, Type{VLines}, Type{HLines}, Type{ABLines}}; ...)` so that legends from `VLines`, `HLines` and `ABLines` also show the styling.
- Created `legend_elements(::Type{ScatterLines}; ...)` to support `ScatterLines` as well.

Intended to close #435

### DIFF
--- a/src/guides/legendelements.jl
+++ b/src/guides/legendelements.jl
@@ -13,6 +13,15 @@ function legend_elements(::Union{Type{Lines}, Type{VLines}, Type{ABLines}};
     return [LineElement(; linecolor=color, kwargs...)]
 end
 
+function legend_elements(::Type{ScatterLines};
+                         marker=from_default_theme(:marker),
+                         markerpoints=[Point2f(0.5, 0.5)],
+                         color=from_default_theme(:markercolor),
+                         kwargs...)
+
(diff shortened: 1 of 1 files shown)


greimel:
It would be nice to come up with a more general solution, but until then I think this is reasonable to do.

greimel (approved):

PR #477, opened 2023-11-22
Set color of minor ticks in default theme

fabern:
Currently AlgebraOfGraphics defines a plot theme. There, the *major* tick lines are changed to `:grey` instead of `:black`. *Minor* ticks are not shown. Their color is not overwritten from Makie default.

I suggest to override also the color of minor ticks from Makie's default `:black` to `:grey`. 
Minor ticks are still deactivated by default. 
However, if now users want to show minor ticks they do not need to change the color.


piever:
Makes sense, thanks for the PR!

piever (approved):

PR #478, opened 2023-11-23
bump version and add deprecation

piever:


src/transformations/visual.jl
```diff
@@ -2,12 +2,14 @@ struct Visual
     plottype::PlotType
     attributes::NamedArguments
 end
-Visual(plottype=Plot{Any}; kwargs...) = Visual(plottype, NamedArguments(kwargs))
+Visual(plottype::PlotType=Plot{Any}; kwargs...) = Visual(plottype, NamedArguments(kwargs))
```
SimonDanisch, line 5: Ah here it is!!! I was searching where this got initialized... 
I think this needs to be Plot{plot} now!

SimonDanisch:
We should test this against https://github.com/MakieOrg/Makie.jl/pull/3401, since it changes plottype a bit, which AoG relies on quite a bit I think?

piever:
Ah I see, yes, `plottype` makes much more sense to me now! With this changes tests pass locally for me on https://github.com/MakieOrg/Makie.jl/pull/3401, I can merge and tag this when Makie 0.20.1 gets released.

PR #486, opened 2024-01-22
Reference `nonnumeric` in doctring of `sorter`

KronosTheLate:
I would have needed this pointer, so I added it when discovering the possibility.

### DIFF
--- a/src/helpers.jl
+++ b/src/helpers.jl
@@ -110,6 +110,11 @@ end
     nonnumeric(x)
 
 Transform `x` into a non numeric type that is printed and sorted in the same way.
+
+# Example
+```
+    mapping(color = :n_children => nonnumeric => "Number of Children")
+```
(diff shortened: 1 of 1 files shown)


asinghvi17 (approved):

jkrumbiegel:
Closing this due to outdated doctests, but I also don't quite see the connection between `sorter` and `nonnumeric`. Any categorical variable will be sorted by default, that's not special about `nonnumeric`?

KronosTheLate:
It is just that I figured I was not the only one who would want to use integers as categorical variables, converted them to strings, and then wanted to sort them and thereby found the `sorter` function. For that situation it would be more apropriate to use `nonnumeric`, which is why I thought it made sende to mention it in the docstring. Also, the examples are usefull in ane case, so closing feels a little premature.

PR #489, opened 2024-01-31
Fix a typo leading to `paginate(plt; row, col)` failures

lcontento:
~~I have a plot with many facets that I am not able to `paginate` using the `col` and `row` arguments.~~
`paginate(plt; row, col)` called on a plot that has no `row`/`col` inside `mapping` throws an error.
Fixing this typo prevents the error.

### DIFF
--- a/src/paginate.jl
+++ b/src/paginate.jl
@@ -75,7 +75,7 @@ function paginate_layer(layer::ProcessedLayer; layout=nothing, row=nothing, col=
         paginate_layer(layer, row_data => row, col_data => col)
     elseif valid(row_data, row)
         paginate_layer(layer, row_data => row)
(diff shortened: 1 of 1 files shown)


lcontento:
After double-checking, while this solves the error that `paginate` raised, it does not fix my actual problem... I will look a bit more into this and update this PR. If I can't get to the bottom of it, then I'll close the PR and open an issue.

lcontento:
In the end, I did not read with enough attention the docstring and did not notice that you need to specify `row` and `col` in `mapping` too... Still, I think `paginate(plt; row, col)` should not throw never throw an error, so fixing this typo is not completely useless even if not particularly useful.

jkrumbiegel (approved):

PR #492, opened 2024-03-04
Support Makie@0.21

asinghvi17:
This PR updates Makie + GridLayoutBase compat, and bumps the patch verion.


asinghvi17:
CC @SimonDanisch

asinghvi17:
The main error appears to be
> │    Invalid attribute cycle for plot type <someplottype>.

asinghvi17:
https://github.com/MakieOrg/AlgebraOfGraphics.jl/blob/7c143ee2dc82ea0cc705498fee97d03c043481cc/src/algebra/layer.jl#L282 

is the line which adds `cycle` to the attributes.  @jkrumbiegel what's the best way to get around this?  As I understand, AoG sets this attribute so that even when plotting into an existing axis, the cycling is ignored.  However, the plotting is done after the step where `Axis` pops the `cycle` attribute.

jkrumbiegel:
Hmm @SimonDanisch the cycle machinery changed since I last looked at it, would you say that it makes sense to allow `cycle` for any plot type because it's one of those generic options that is not a "real" attribute? In principle you may cycle any attribute you please, so just because a plot object doesn't cycle anything in their theme, that doesn't mean adding the option should error I think.

SimonDanisch:
> would you say that it makes sense to allow cycle for any plot type because it's one of those generic options

Does it work generically for any recipe, though? Do we through a descriptive error, if we don't find a palette to cycle through?
If yes, I think we could do that.. 
How are we documenting it then? Put it into the general attributes with some docs?
Although I dont think we add all generic attributes to every recipe (I think we maybe should, right?)

jkrumbiegel:
It should work generically, yes. It just looks up attributes by key in the linked palettes, you can use any keys in either of them. Not sure about the errors, too long ago that I wrote this. I'm just saying, we could either add empty cycles to all recipes that don't have any predefined (to avoid them erroring with AoG), or we add `cycle` as an allow-list attribute.

jkrumbiegel:
Looks like there might be a Makie bug here regarding the handling of `PlotList` and cycling. Not quite sure why, this is the MWE

```julia
p = Choropleth(
    ([Polygon([Point(0, 0), Point(1, 0), Point(1, 1)])],),
    Dict{Symbol, Any}(:color => [1], :cycle => nothing, :colorrange => (1, 1))
)
f = Figure()
a = Axis(f[1, 1])
plot!(a, p)
f
```

Interestingly,

```julia
p = Choropleth(
    ([Polygon([Point(0, 0), Point(1, 0), Point(1, 1)])],),
    Dict{Symbol, Any}(:color => [1], :cycle => nothing, :colorrange => (1, 1))
)
```

gives

```
┌ Warning: PlotSpec objects are supposed to be title case. Found: poly. Please use Poly instead.
└ @ Makie ~/.julia/packages/Makie/We6MY/src/specapi.jl:36
Plot{Makie.plotlist, Tuple{PlotSpec}}
```

and 

```julia
p = Scatter(
    ([1, 2, 3],),
    Dict{Symbol, Any}(:color => [1, 2, 3], :cycle => nothing,)
)
```

gives 

```
Scatter{Tuple{Vector{Point{2, Float64}}}}
```

so maybe it's about that somehow

PR #499, opened 2024-05-18
add Dictionaries@0.4 to compat

aplavin:
Shouldn't require any code changes I think


hv10:
This would be very nice to have as the Dictionaries@0.3 requirement is blocking the use of AOG for me - as it leads to version downgrades for a lot of other packages (e.g. RxInfer.jl)

greimel:
Thanks!

Issue #262, opened 2021-08-17
How to create an ordered bar plot?


roland-KA:
I'm trying to create a bar plot which shows the population size of different regions and would like to order the bars according to population size (see: https://github.com/roland-KA/StatisticalPlotsWithJulia, the bar plot is in the Pluto notebook `notebooks/DV-Basics-AlgebraOfGraphics.jl` in section "*Population by Subregion*") 

But up to now, I didn't succeed:
- AoG ignores the sorting order of the DataFrame given as an argument to `data`. So sorting the data in advance doesn't help.
- Using `sorter` changes the order of the bars, but not in the way I intended. This is perhaps a problem of the documentation (and my lack of understanding), which doesn't tell much about what `sorter` really does.

A detailed description of the problem as well as the code I created so far can be found in the Pluto notebook mentioned above.

AugustoCL:
I'm facing the same problem that AoG ignores the sorting order of the DataFrame. 

I would like to let here a snippet of a reproducible code of that case, to help the devs.

```
import Downloads: download
using CSV, DataFramesMeta, Statistics
using AlgebraOfGraphics, CairoMakie

# download and read data
const DATADIR = "netflix/data/";
filename = DATADIR * "netflix_titles.csv"
download("https://raw.githubusercontent.com/rfordatascience/tidytuesday/master/data/2021/2021-04-20/netflix_titles.csv",
filename)
rawdata = CSV.read(DATADIR*"netflix_titles.csv", DataFrame)

# clean raw data
df = @chain rawdata begin 
    transform!(
        :date_added => ByRow(passmissing(x -> Date(strip(x), dateformat"U d, Y"))) => :date_added_clean,
        :duration => ByRow(x -> split(x, " ")) => [:durationtime, :durationunits])
    transform!(:durationtime => x->parse.(Int,x), renamecols=false)
end

# median duration per genre
dfp = @chain df begin
    subset(
        :durationunits => (x->x .== "min"),
        :type => (x-> x.=="Movie"))
    transform(:listed_in => ByRow(x-> split(x, ", ")) => :genre)
    flatten(:genre)
    groupby(:genre)
    combine(:durationtime => median => :mediantime)
    sort(:mediantime, rev=true)
end

plt = data(dfp) * 
    mapping(:genre => "Median Duration time", :mediantime => "") * 
    visual(BarPlot, direction=:x, color=:darkblue)
f = Figure()
draw!(f, plt; 
[...]

PR #529, opened 2024-08-23
Presorted data

jkrumbiegel:
Fixes https://github.com/MakieOrg/AlgebraOfGraphics.jl/issues/262

This PR introduces a new `mapping` helper called `presorted`. If this is applied to a column, AoG will internally keep track of the original order of appearance of these categorical values. One could of course require users to always use CategoricalArrays, but this is an annoying overhead for exploratory plotting.

## Example

A dataframe which contains some ordered data about countries.

```julia
julia> 
       countries = ["Algeria", "Bolivia", "China", "Ecuador", "France", "Germany", "Honduras"]
       group = ["2", "3", "1", "1", "2", "3", "1"]
       some_value = exp.(sin.(1:7))

       df = DataFrame(; countries, group, some_value)
       sort!(df, :some_value)
7×3 DataFrame
 Row │ countries  group   some_value 
     │ String     String  Float64    
─────┼───────────────────────────────
   1 │ France     2         0.383305
   2 │ Ecuador    1         0.469164
   3 │ Germany    3         0.756226
   4 │ China      1         1.15156
   5 │ Honduras   1         1.92897
   6 │ Algeria    2         2.31978
   7 │ Bolivia    3         2.48258
```

A normal plot will order countries alphabetically:

```julia
data(df) * mapping(:countries, :some_value, color = :group) * visual(BarPlot, direction = :x) |> draw
```

image

If we apply `presorted` to `:countries`, the original order is retained for this column, but not for `:group` which is still reordered:

```julia
data(df) * mapping(:countries => presorted, :some_value, color = :group) * visual(BarPlot, direction = :x) |> draw
```

image

We can apply `presorted` to `:group` as well to retain the original order in the color legend:

```julia
data(df) * mapping(:countries => presorted, :some_value, color = :group => presorted) * visual(BarPlot, direction = :x) |> draw
```

image


jkrumbiegel:
Some notes here from the development process, to refer back to later:

- continuous data is already used in the original order (for example when dealing with lines)
- categorical data is getting sorted however, during the creation of the categorical scales (mergesorted)
- this sorting depends on the category value itself and may not be what the user wants
  - for example, Afghanistan Bolivia China will always be sorted ABC in a plot of area, even if China, Afghanistan Bolivia would be the right order
- a means to disable sorting selectively must be introduced
- a global disabling doesn't make sense, for example if we have countries ordered by size, but they should also be colored by categories A,B,C then we want A,B,C in the legend even if the first country is category B and the second A
- one could "mark" a column in the mapping with a function, for example `presorted`, which wraps all values and changes the sorting order. Because function application is per-element and not whole-column, this needs extra logic if this kind of wrapper type like `Presorted` is encountered, so the original positions of the elements can be written into the wrappers.
  - a problem with this arises when multiple layers with partial data are plotted on top of each other, for example the first plots c,b,d and the other b,a,e. When merging the values for the scale, we'd have c1,b2,d3 for the first and then b1,a2,e3 for the other. The order I'd expect is c,b,d,a,e so b2 and b1 need to be considered the same for this. So maybe == for these values needs to ignore the index from the selection phase? But it will actually order c1,b2,a2,d3,e3 because it doesn't know that these are from different layers. This gets complicated quite quickly actually, and should probably be solved by using categorical values in the first place that carry the real ordering with them. What we want here is more of a quick solution for simple plots.

PR #554, opened 2024-09-06
gallery cards are missing

lazarusA:
@jkrumbiegel  this is the branch. Only the gallery cards were not working. All other things should be fine.


jkrumbiegel:
Hehe that's a big "only", that democards package is probably more work to convert to vitepress because it writes its own html, with its own theming css etc. It seems to me that we'd need to convert those scripts to normal pages with `@example` blocks and then do something similar to Makie to make a page with preview images.

lazarusA:
at some point I did try the new OhMyCards package, with some partial success, but at the time it was really hard for me to incorporate into this current workflow (and specially have author badge show at the top). Probably, things are better now, don't know, will need to check again.

lazarusA:
@asinghvi17 is using it with some success here: https://geo.makie.org/v0.7.4/examples . Maybe later, I could take a look again.

asinghvi17:
I have some time today so will take a look!

jkrumbiegel:
docs are switched to vitepress now

PR #558, opened 2024-09-13
Generic dodging mechanism

### LINKED ISSUE #393: Dodge in scatter,lines
## Problem description
I often encountered the issue, that there is no way to dodge datapoints in a scatter (or less important imho line) plot.

That is going from here:
image
To here: 
image

Something similar is asked here #230 I think

## Proposed solution
I dont understand enough of the internals, but as I understand Makie.jl implements the dodge feature for violin & histograms.

In ggplot2 dodging is possible by specifing a position-function (`position = position_dodge(width=0.2)`). I was expecting something like
`data(df) * mapping(x,y,color="categoryA",dodge="categoryA")  * visual(Scatter)` to "just work", by doding the colored datapoints.

Let's just hope I missed something ;-)

jkrumbiegel:
Fixes #393 

This PR adds the keywords `xdodge` and `ydodge` to `mapping` which, like other hardcoded mappings (`layout`, `row`, `col`, `group`) are not tied to any Makie attributes in particular. Instead, these modify any `AesX` or `AesY` columns, respectively, when they are used. This allows generic dodging of things like scatters or errorbars:

```julia
df = (
    x = [1, 1, 2, 2],
    y = 1:4,
    err = [0.5, 0.4, 0.7, 0.6],
    group = ["A", "B", "A", "B"],
)

data(df) *
    (
        mapping(:x, :y, :err, xdodge = :group, color = :group) * visual(Errorbars) +
        mapping(:x, :y, xdodge = :group, color = :group) * visual(Scatter)
    ) |> draw(scales(DodgeX = (; width = 0.2)))
```

image

The width of the dodging can be controlled with a scale attribute:

```julia
mapping(
    repeat(string.('A' .+ (0:9)), inner = 20),
    randn(200),
    xdodge = repeat(["X", "Y"], 100),
    strokecolor = repeat(["X", "Y"], 100),
) * visual(Scatter, strokewidth = 2, color = :transparent) |>
    draw(scales(DodgeX = (; width = 0.5)))
```

image

# Caveats / remaining design problems

The most common scenario is errorbars on barplots, I think. However, barplots implement their own dodging via the `dodge` keyword. (There are a couple recipes that have `dodge` implemented, rainclouds, crossbars, violins...) And the width does not depend on the dodging scale but on the plot's own `width` setting as well as `gap` and `dodge_gap`. So one can have mismatched dodging:

```julia
df = (
    x = [1, 1, 2, 2],
    y = 1:4,
    err = [0.5, 0.4, 0.7, 0.6],
    group = ["A", "B", "A", "B"],
)

data(df) *
    (
        mapping(:x, :y, dodge = :group, color = :group) * visual(BarPlot) +
        mapping(:x, :y, :err, xdodge = :group) * visual(Errorbars)
    ) |> draw(scales(DodgeX = (; width = 0.2)))
```

image

One could use `xdodge` for barplot as well, but then this can conflict with the `width` setting from barplot, which would then not be adjusted for anymore (barplot wouldn't know that dodging is going on)

```julia
df = (
    x = [1, 1, 2, 2],
    y = 1:4,
    err = [0.5, 0.4, 0.7, 0.6],
    group = ["A", "B", "A", "B"],
)

data(df) *
    (
        mapping(:x, :y, xdodge = :group, color = :group) * visual(BarPlot, width = 1) +
        mapping(:x, :y, :err, xdodge = :group) * visual(Errorbars)
    ) |> draw(scales(DodgeX = (; width = 0.2)))
```

image

Note that in this particular case, if `width = 1` wasn't set, the plot would look ok, but that might not always be the case if enough categories are missing, so it's too brittle I think.

image

So ideally, I think the `width` for widthless plots (scatter, errorbars, etc.) could be inferred from plots with width that are also using the dodge scale. So if a barplot is in the mix, the width would be picked up from it somehow. This adds a small layer of complexity on top of the processing pipeline but might be manageable.


jkrumbiegel:
Going the width inference route now, to make the super common case of barplots + errorbars easier. The widthless errorbars can now infer the necessary width from the barplots in a sort of preprocessing step before entries are computed.

```julia
df = (
    x = [1, 1, 2, 2],
    y = 1:4,
    err = [0.5, 0.4, 0.7, 0.6],
    group = ["A", "B", "A", "C"],
)

data(df) *
    (
        mapping(:x, :y, dodge = :group, color = :group) * visual(BarPlot, gap = 0.5) +
        mapping(:x, :y, :err, dodge_x = :group) * visual(Errorbars)
    ) |> draw
```
image

PR #564, opened 2024-09-19
Enforce that all axis types are <: AbstractAxis

asinghvi17:
As brought up in Slack just now.  Still needs tests but works locally with GeoAxis.

One question that remains is what to do about polar axes.  With the naive implementation here, `draw(; axis = (; type = PolarAxis))` fails because the constructor tries to pass `xlabel` etc in the kwargs, which PolarAxis can't handle.

### DIFF
--- a/src/entries.jl
+++ b/src/entries.jl
@@ -27,7 +27,7 @@ end
 # Use technique from https://github.com/MakieOrg/AlgebraOfGraphics.jl/pull/289
 # to encode all axis information without creating the axis.
 struct AxisSpec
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
xlabel and ylabel just happen to be the names for dim1 and dim2 in Axis, for PolarAxis it's different. So maybe instead of translating from `xlabel` a previous step would just pick the correct label for a different type. One thing that will also be a bit wrong is that the aesthetics are called AesX and AesY. So then you'd have to do `scales(X = ...)` also for PolarAxis

PR #572, opened 2024-10-12
Add stephist and related Stairs plot type

Beforerr:


Beforerr:
Hi @jkrumbiegel , I am just wondering  is there any reason stop us from supporting Volume type?

jkrumbiegel:
No reason, just that Volume wasn't a priority for me personally. But in the long run AoG should support most Makie recipes

PR #592, opened 2025-01-27
Break `mergeable` up so that it's extensible

asinghvi17:
Breaks up `mergeable(::ProcessedLayer)` into `mergeable(layer.plottype, layer.primary)` for dispatch purposes.  This should enable `mergeable` to be extended by other packages that have recipes.


asinghvi17:
The doc build is failing because it wants `mergeable` to be documented somewhere.  Seeing as we don't have a public API doc page for overloadable functions for custom recipes, I may as well write one...

PR #599, opened 2025-02-12
do dark mode logo

lazarusA:
- this adds some css to include the logo when we are in the dark mode. 
- limits the max-size of the `demo_hero` image, which currently is too big in big screens 😄 .

the logo at the top left should be fixed once https://github.com/LuxDL/DocumenterVitepress.jl/issues/208 is solved 😓 .

### DIFF
--- a/docs/src/.vitepress/theme/style.css
+++ b/docs/src/.vitepress/theme/style.css
@@ -211,3 +211,7 @@
       user-select: none;
       margin: 0 0 8px;
   }
+  
+  .dark .VPHero .image-src {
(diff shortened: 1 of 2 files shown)


jkrumbiegel:
@lazarusA you can also do PRs within the repo I think? Then the previews would build :)

lazarusA:
> @lazarusA you can also do PRs within the repo I think? Then the previews would deploy :)

can I? last time I tried, I didn't have access to rights 😄 .

jkrumbiegel:
I see, member of MakieOrg only means read access. I've added you

PR #609, opened 2025-02-26
Draw to Makie.SpecApi spec

jkrumbiegel:
Turning an AoG spec into a Makie spec allows to update an AoG plot interactively

https://github.com/user-attachments/assets/047c72eb-4620-44a0-b3e9-d0b6f9038c1c

This is so far only a very basic proof of concept.

### DIFF
--- a/src/draw.jl
+++ b/src/draw.jl
@@ -255,3 +255,43 @@ function Base.showerror(io::IO, pe::PaletteError)
         """
     print(io, msg)
 end
+
+function draw_to_spec(spec, scales = scales())
+    agrid = compute_axes_grid(spec, scales)
+
+    S = Makie.SpecApi
+
+    axisspecs = map(vec(agrid)) do ase
+        ax = ase.axis
+        axtype = ax.type === Axis ? S.Axis : ax.type === Axis3 ? S.Axis3 : error()
+
+        isempty(ase.entries) && return nothing
+
+        plots::Vector{Makie.PlotSpec} = map(ase.entries) do entry
+            Makie.PlotSpec(entry.plottype, entry.positional...; pairs(entry.named)...)
+        end
+
+        ax.position => axtype(; plots, pairs(ax.attributes)...)
+    end
+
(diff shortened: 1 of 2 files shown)


ajinkya-k:
@jkrumbiegel can you share the code you used to make the demo video in the PR description?

jkrumbiegel:
You didn't plot the spec observable into the figure, that's the missing piece

ajinkya-k:
Sorry I’m getting lost in the types here. Do I plot the spec with `plot!(fig, spec)`

ajinkya-k:
I tried `plot!(ax, spec)` iside the record block where `ax = Axis(f[1,1])` but i get the following error: 
```
ERROR: MethodError: no method matching draw_atomic(::GLMakie.Screen{GLFW.Window}, ::Scene, ::Plot{MakieCore.plot, Tuple{Makie.GridLayoutSpec}})
The function `draw_atomic` exists, but no method is defined for this combination of argument types.
```

ajinkya-k:
same code with CairoMakie gives: 
```
┌ Warning: Plot{MakieCore.plot, Tuple{Makie.GridLayoutSpec}} is not supported by cairo right now
└ @ CairoMakie ~/.julia/packages/CairoMakie/Ji9yW/src/infrastructure.jl:186
```

ajinkya-k:
Interestingly using "current_figure()" works with GLMakie: 
```
record(current_figure(), "color_animation.mp4", r_iterator; framerate = 20) do r
    println(r)
    spec[] = AlgebraOfGraphics.draw_to_spec(drawnb(r))
    plot(spec)
end
```

SimonDanisch:
Could we merge this as an experimental feature? :)

jkrumbiegel:
@SimonDanisch yeah, if you could take it for a spin and see if the axis linking works, and if not maybe try to fix it? Without that it's not really functional enough I think for interactive use. Other than that I think it's good enough by now to be an experimental addition.

PR #617, opened 2025-03-17
added informative error when legend is not a named tuple

### LINKED ISSUE #616: Bad error message when using `legend` keyword for `draw`
I was trying to change the position of the `legend` by doing something like this: 
```
draw(... legend = (position = :bottom))
```
and got the error (see at the end). This error is hard to parse but after some experimentation I figured out either of the following works:
```
draw(... legend = (position = :bottom, titleposition=:left)) 
draw(... legend = (position = :bottom, ))  # the `,` in the end with nothing after that is crucial
```

So I guessed that the `legend` keyword must have (a) alteast two name-value pairs or (b) a hanging "," to force it to be a NamedTuple, otherwise it collapses to a "scalar" (in this case a Symbol) but the error does not say that. Is it possible to make the error message better? This is similar to #575 

```julia-repl
MethodError: no method matching iterate(::Symbol)
The function `iterate` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  iterate(::Layers, Any...)
   @ AlgebraOfGraphics ~/.julia/packages/AlgebraOfGraphics/E0gIq/src/algebra/layers.jl:16
  iterate(::LaTeXStrings.LaTeXString, ::Int64)
   @ LaTeXStrings ~/.julia/packages/LaTeXStrings/6NrIG/src/LaTeXStrings.jl:108
  iterate(::LaTeXStrings.LaTeXString)
   @ LaTeXStrings ~/.julia/packages/LaTeXStrings/6NrIG/src/LaTeXStrings.jl:109
  ...


Stacktrace:
 [1] isempty(itr::Symbol)
   @ Base ./essentials.jl:1121
 [2] _kwdict(prs::Symbol)
   @ AlgebraOfGraphics ~/.julia/packages/AlgebraOfGraphics/E0gIq/src/draw.jl:33
 [3] draw(d::Layer, scales::AlgebraOfGraphics.Scales; axis::@NamedTuple{}, figure::@NamedTuple{size::Tuple{Int64, Int64}}, facet::@NamedTuple{}, legend::Symbol, colorbar::@NamedTuple{}, palette::Nothing)
   @ AlgebraOfGraphics ~/.julia/packages/AlgebraOfGraphics/E0gIq/src/draw.jl:82
 [4] top-level scope
   @ In[32]:3
```

ajinkya-k:
the `draw` function calls `_kwdict` on `legend` argument, which expects a `NamedTuple` or at least a `Tuple`. If only one name-value pair is provided such as `draw(..., legend = (position = :bottom))` the legend is of type `Symbol` and not a named tuple. The error thrown when this happens is uinformative, see #616. This PR provides a more helpful error with a suggestion and tells the use to add a hanging comma like here: `draw(..., legend = (position = :bottom, ))`

fixes #616

### DIFF
--- a/src/draw.jl
+++ b/src/draw.jl
@@ -79,7 +79,7 @@ function draw(d::AbstractDrawable, scales::Scales = scales();
     axis = _kwdict(axis)
     figure = _kwdict(figure)
     facet = _kwdict(facet)
-    legend = _kwdict(legend)
+    legend = (!isa(legend, NamedTuple)) ? throw(DomainError("legend provided is not a NamedTuple. If you specified a single name-value pair, add a hanging comma like so: (name = value, )")) : _kwdict(legend)
     colorbar = _kwdict(colorbar)
 
     return _draw(d, scales; axis, figure, facet, legend, colorbar)
--- a/test/legend.jl
+++ b/test/legend.jl
@@ -155,3 +155,10 @@ end
     @test Makie.to_value(els[3].alpha) == 0.5
     @test Makie.to_value(els[4].alpha) == 1.0
 end
+
+@testset "legend named tuple #PR #617" begin
+    dat = DataFrame(grp = ["A", "B", "C"], x = [1., 2., 3.], y = [1., 2., 3.])
+    plt = data(dat) * mapping(:x, :y, color = :grp) * visual(Scatter)
+    @test_throws DomainError draw(plt, legend = (position = :bottom))
+    @test isa(draw(plt, legend = (position = :bottom,)), AlgebraOfGraphics.FigureGrid)
+end


jkrumbiegel:
It expects any key value object that has Symbol keys, so NamedTuple is just one of the possibilities. I agree that we could have a better error message if conversion to the dict doesn't work. Not sure what's the most robust way to do that given that we want to have a nice error. Maybe try catch around the current t conversion is easiest, with a nice error repeating the input.

ajinkya-k:
I see. I started this PR just as a proposal, and we can iterate on it.

ajinkya-k:
I think it has to be at the tope level of `draw` function call because we should be able to inform the user which argument failed right?

jkrumbiegel:
You can just pass a symbol into the conversion function as well and splice it into the error. Like `_kwdict(x, kind = :legend)`

jkrumbiegel:
superseded

PR #641, opened 2025-04-04
Add `AesLineWidth` and extend the mappings understood by `VLines`

andreuvall:
On the one hand, the PR adds the `AesLineWidth` aesthetic, which can be used by different `Visual`s. On the other hand, the `VLines` visual is modified to understand both `:linestyle` and `:linewidth` mappings, allowing for plots like the following:
```julia
using DataFrames
using AlgebraOfGraphics
using CairoMakie

df = DataFrame(t = [1, 2], x = ["A", "B"])
draw(data(df) * mapping(:t; linewidth = :t, linestyle = :x) * visual(VLines))
```
image


jkrumbiegel:
Thanks for this! I've continued it in https://github.com/MakieOrg/AlgebraOfGraphics.jl/pull/656 so I'm going to close this one

andreuvall:
Thank you, @jkrumbiegel. I definitely wanted to continue this PR, but I got tests failing on previous Julia versions. I would have needed to take a deeper look, but then other priorities kept me away from it.

PR #642, opened 2025-04-06
Remove PR doc previews for closed and merged PRs

abhro:


jkrumbiegel:
thanks, I wonder why these weren't deleted automatically, there should be a github workflow for cleanup

abhro:
I saw that too haha. I've noticed that workflow can fail for various reasons. Since there are workflow failures and also many CompatHelper branches, it's probably worth taking a look

jkrumbiegel:
I've updated the workflow with the code from Makie, that one works better

PR #698, opened 2025-11-16
expand docs for wide vs long & mapping arguments

ericphanson:
Tries to address https://discourse.julialang.org/t/algebraofgraphics-documentation-frustrations/133903 using @jkrumbiegel's comments/examples from https://discourse.julialang.org/t/algebraofgraphics-documentation-frustrations/133903/2.

Written with the help of Claude code (but checked/modified by me).

### DIFF
--- a/docs/src/philosophy.md
+++ b/docs/src/philosophy.md
@@ -156,7 +156,7 @@ format is sometimes more natural for your data source.
 This flexibility is achieved in three ways. First, the
 [Tables interface](https://github.com/JuliaData/Tables.jl)
 ensures integration with a large variety of data sources.
-Second, using [Wide data](@ref "Wide-data") syntax with multidimensional mappings, users can plot
+Second, using wide data syntax with multidimensional mappings, users can plot
 many columns together without first reshaping to long format.
 Finally, tabular datasets are not a requirement: users may also work directly
 with [Pre-grouped data](@ref "Pre-grouped-data"), which are not organized as a table,


docs/src/philosophy.md
```diff
@@ -141,18 +141,26 @@ and a
...
+format leads to simpler AlgebraOfGraphics code (mappings are just column symbols), wide
+format is sometimes more natural for your data source.
+
+This flexibility is achieved in three ways. First, the
 [Tables interface](https://github.com/JuliaData/Tables.jl)
 ensures integration with a large variety of data sources.
-Second, using the TODO REFLINK Wide data syntax, users can compare many different
-columns in the same visualization, without having to first reshape the dataset
-to a long format.
+Second, using [Wide data](@ref) syntax with multidimensional mappings, users can plot
```
ericphanson: this actually points to `tutorials/intro-v#Wide-data` since I renamed the title in `docs/src/examples/data-manipulations/wide-data.md`. But I thought it was a decent example to point to still, and we point to `docs/src/examples/data-manipulations/wide-data.md` below.

Eben60:
In the introduction, I'd suggest a change as following

No familiarity with plotting or data analysis in Julia is required but basic familiarity with the Julia language 
**and a few packages ([DataFrames.jl](https://github.com/JuliaData/DataFrames.jl), anything else?**..)
is expected such that you can read and understand the code.
**A general familiarity with Makie.jl would be an advantage.**

Issue #366, opened 2022-03-02
Confidence Interval Band


kir0ul:
## Problem description

I would like to be able to plot a line with a confidence interval band, something like the plot below (which I believe is not possible at the moment using only `AlgebraOfGraphics.jl` alone).

![image](https://user-images.githubusercontent.com/6053592/156365743-03de79ac-009a-4874-bea7-b3e1ef634824.png)

## Proposed solution

I guess the implementation could use the [`band()` function from Makie](https://makie.juliaplots.org/stable/examples/plotting_functions/band/index.html), and borrow some ideas for the API from [VegaLite.jl](https://www.queryverse.org/VegaLite.jl/stable/examples/examples_error_bars_bands/#Line-Chart-with-Confidence-Interval-Band-1) or [ggplot2](https://ggplot2.tidyverse.org/reference/geom_ribbon.html).

greimel:
`linefill()` does that http://juliaplots.org/AlgebraOfGraphics.jl/stable/API/recipes/#AlgebraOfGraphics.linesfill 

Unfortunately it is not covered in the docs yet.

kir0ul:
Thanks! I've been able to use it on a single line plot but I'm struggling to use it on a plot with multiple lines and keep the same colors.
Would you have any hint or example on how to use `linesfill()` with multiple line using `AlgebraOfGraphics` syntax? Or maybe I should switch to plain `Makie` syntax?

I need to do something like the figure below.
![image](https://user-images.githubusercontent.com/6053592/158481447-cdbeef9c-d1a5-4dae-b109-d31797506c13.png)

And for now my code is looking like the following:
```julia
    plt =
        data(df_avg) *
        mapping(
            :episode => "Episode",
            :reward_avg => "Cummulated rewards",
            color = :γ => "γ",
        ) *
        visual(Lines)
    fg = draw(plt; axis = (; title = title), figure = (; backgroundcolor = :transparent))
    for (i, d) in enumerate(dicts)
        linesfill!(
            collect(1:n_episodes),
            vec(df_avg.reward_avg[df_avg.γ.==d["γ"], :]);
            lower = lower[i, :],
            upper = upper[i, :],
        ) 
    end
```

greimel:
Can you add some fake data to make your example runnable? Then it will be easier to help.

Nosferican:
[...]

PR #710, opened 2025-12-07
feat: Confidence interval bands + errorbars using GLM.jl v1

icweaver:
Weighted `linear` fits now display confidence-interval bands. Previously, passing `weights` via `mapping` silently disabled the uncertainty band, because weighted `GLM.lm` does not support confidence intervals (https://github.com/JuliaStats/GLM.jl/issues/495).

Weighted fits are now computed with `GLM.glm`, which does support them, so weighted and unweighted fits both default to a 95% confidence band.

```julia
using AlgebraOfGraphics, CairoMakie, Random

df = let
    x = 1:5
    y_err = randn(Xoshiro(111), length(x))
    y = x .+ y_err
    y_unc = abs.(y_err)
    (; x, y, y_unc)
end

data(df) * mapping(:x, :y) * (
    (visual(Scatter) + mapping(:y_unc) * visual(Errorbars)) * visual(; label = "data")
    + mapping(; weights = :y_unc) * linear() * visual(; color = Makie.wong_colors()[1], label = "model")
) |> draw
```

## Before

Weighted linear fit without confidence band

## After

Weighted linear fit with confidence band

## New keywords

`linear` gains three keywords:

- `weighttype = :fweights`: the `StatsBase.AbstractWeights` type used to interpret the `weights` column. Currently only `:fweights` is accepted: GLM v1 treats every weight kind as frequency weights, so accepting the others would silently compute frequency-weight statistics. `:aweights` and `:pweights` will be enabled in a follow-up PR once GLM v2 is released.
- `distr = GLM.Normal()`: distribution passed through to `GLM.glm` for weighted fits.
- `link = GLM.canonicallink(distr)`: link function passed through to `GLM.glm` for weighted fits.

Design note: AoG [strips array wrappers](https://github.com/MakieOrg/AlgebraOfGraphics.jl/pull/710#discussion_r2734354720) at the layer-processing level, so a user-constructed `aweights(...)` column would arrive at the analysis as a plain `Vector`. This is why the weight kind is specified via a keyword harness further down the pipeline rather than being carried by the vector type. Whether to preserve wrapper types instead is left to the GLM v2 follow-up (see review discussion).

## Other changes

- `interval = :prediction` is not available for weighted fits, as GLM only defines confidence intervals for `glm` models. This is noted in the docstring.
- Unit tests cover the weighted band values against `GLM.predict(...; interval = :confidence)` and the error thrown for unsupported `weighttype` values.
- GLM compat bumped from 1.9.1 to 1.9.2.

## Related

- https://github.com/MakieOrg/AlgebraOfGraphics.jl/issues/366

## Future work (GLM v2 follow-up)

- Enable `weighttype = :aweights` and `:pweights` (proper support from https://github.com/JuliaStats/GLM.jl/pull/487), and revisit whether the weight kind should be carried by the `AbstractWeights` vector type instead of a keyword.


icweaver:
Ok, I think things should be ready now for a first review

docs/src/reference/analyses.md
```diff
@@ -152,6 +152,34 @@ specs = data(df) * mapping(:x, :y, color=:a => nonnumeric) * (
...
+    x = 1:5
+    y_err = randn(length(x))
+    y = x .+ y_err
+    y_unc = abs.(y_err)
+    (; x, y, y_unc)
+end
+specs = data(df) * mapping(:x, :y) * (
+    (visual(Scatter) + mapping(:y_unc) * visual(Errorbars)) * visual(; label = "data") +
+    mapping(; weights = :y_unc) * (
+        linear(; weighttype = :fweights, weighttransform = x -> inv.(x .^ 2)) * visual(; color = colors[1], label = "fweights")
```
nalimilan: Is `weighttype` really needed? The whole idea behind these `AbstractWeights` types in StatsBase is that you only specify once the kind of weights you have when building the vector, and then all functions automatically interpret it correctly. We don't have any API currently that takes the kind of weight separately.

src/transformations/linear.jl
```diff
@@ -50,7 +81,7 @@ function (l::LinearAnalysis)(input::ProcessedLayer)
 end
 
 """
-    linear(; interval=automatic, level=0.95, dropcollinear=false, npoints=200)
+    linear(; interval=automatic, level=0.95, dropcollinear=false, npoints=200, weighttype=:fweights, weighttransform=identity, distr=GLM.Normal())
```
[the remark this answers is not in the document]
jkrumbiegel: I didn't closely examine this but in general, I like it better if options for a different package are pass-through (like `kwargs...)`, so AoG doesn't really add API on its own. But if it's just really cumbersome to use that way, that's an argument for making special AoG API. After all, `linear()` is still supposed to be rather simple to use.

docs/src/reference/analyses.md
nalimilan: My point is that imagine I'm loading a dataset in `df`. I immediately do something like `df.y_unc = aweights(df.y_unc)`. Then I can create one or more graphs, estimate models, etc., without ever saying again that these are `aweights`, and everything works automatically.  And this is necessary to use weights with StatsBase or GLM as they require an `AbstractWeights` type. Introducing a new keyword argument which differs from the method used in other packages actually makes things more complex IMO.

As @jkrumbiegel said (https://github.com/MakieOrg/AlgebraOfGraphics.jl/pull/710/changes#r2685551356) it seems better to adopt the API of underlying packages as much as possible.
without introducing new API in AoG.

icweaver: Right, so your strategy is totally what I would do if I was using vanilla Makie.

In contrast, it's my understanding that the current usage with AoG makes all calls to StatsBase and GLM internally so that the user only needs to pass the raw data without any pre-processing or explicit loading of additional packages needed. It seems this would need to change with your proposed usage, so I guess my question is if this is the direction that we would like to go in this PR.

At any rate, happy to go with either route, was just asking for clarity

src/transformations/linear.jl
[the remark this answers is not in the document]
icweaver: Hi all, I just added a simple pass-through kwarg like Jules recommended. Re-tooling `linear` to accept positional args sounds like it could be a breaking change that might be better for a separate PR? Anyway, I've updated the examples above for your review using this simple setup now.

docs/src/reference/analyses.md
nalimilan: I'm not familiar enough with AoG to give an informed opinion on the implementation, but from an external API point of view it would be too bad IMO that users would need to repeat the weights type as the `weighttype` argument given that it's already carried by the vector type.

jkrumbiegel: > I guess this could be selectively patched so that calls to f(identity, vs) could be strategically avoided

I don't know, you could try to pass through the original data if the transformation function is `identity`. I can't say without trying if that would have weird side effects. But in general, AoG really only cares about the eltypes because it assumes that the vector types are side effects of whatever table format is being used. Like why should it make a difference if a vector, a vector view, a lazy mapped array, etc. is used if they all contain floats? In that sense, `AnalyticWeights` is the odd one here as it doesn't contain `Weight` types but normal numbers, and therefore disintegrates into normal `Vector{T}` under transformations. I'm not sure making AoG sensitive to wrapper types is a good behavior change.

The `map(identity, x)` behavior and changing of array types through transformations is of course a bit of a Julia footgun sometimes, and we also have special cases for that in Makie. You have a transformation function with a default of "no transformation" (== identity) but don't want to pay for an actual copy so you special case that `map(identity, x)` stays just `x`.

icweaver:
Hey folks, I cleaned things up a bit here and tightened the scope to GLM.jl v1. The PR description is current now for your review

PR #720, opened 2026-01-08
fix `presorted` not working with color scales

TheLostLambda:
The newly added reference test used to throw a runtime error — it now produces the expected plot.

https://github.com/MakieOrg/AlgebraOfGraphics.jl/blob/c2751aa2ae907a642b97d64fa5a7802aa572dfbc/src/scales.jl#L88 shows the code that caused the failure. The `Presorted("key")` would not compare equal to the `"key"` in the color mapping.

This feels like a reasonable way for `isequal(Presorted)` to behave, as it's in line with the existing definition that ignores the `.i` field.

This is my first PR to this repo and one of my first to Julia, so let me know if I've missed anything!


TheLostLambda:
Perhaps there should be an abstract type for these wrappers, then all of the current wrappers can be a subtype of that?

That way users could also add their own wrappers and just "inherit" from that abstract wrapper type so that they are automatically unwrapped too?

jkrumbiegel:
I'm usually not a fan of abstract types because they only work for one thing, while simply defining methods / traits can be used for however many behaviors.

jkrumbiegel:
closing this in favor of #766

PR #751, opened 2026-04-28
Add aesthetics for stem plot

stelmo:
Add basic functionality for `Stem` aesthetics and legend. Should I make a test for it?


stelmo:
I came across this problem when making a test for my PR: the stem gets split apart if the legend items cleanly divide the plot into parts, see below. Do you have any clues on how to prevent this behaviour?
```julia
df = (;
    x = [1, 2, 3, 4],
    y = [1, 2, 0.5, 1.3],
    color = ["a", "b", "a", "b"],
)
spec = data(df) * mapping(:x, :y, color = :color) * visual(Stem)
draw(spec, legend = (; hide_unused = true))
```
image

```julia
df = (;
    x = [1, 2, 3, 4],
    y = [1, 2, 0.5, 1.3],
    color = ["a", "a", "b", "b"], # <= changed the order here
)
spec = data(df) * mapping(:x, :y, color = :color) * visual(Stem)
draw(spec, legend = (; hide_unused = true))
```
image

Also, I am not sure why the other 2 tests are failing, I did not touch that code.. Maybe I pulled in a bug from master?

jkrumbiegel:
By default AoG creates a separate plot object for each group. If that's not desired for some plot type, they have to be merged together explicitly. It's done for barplot for example so that the bar widths are not calculated in separation. It's not the cleanest logic I think because not every attribute is array-compatible but have a look if you can apply that to stem. Search for merge

stelmo:
@jkrumbiegel I think this PR is done. I made a test using all the features I implemented (called `stem plot`). The failing tests are a mystery, as they involve unrelated functions (`contour analysis` and `title subtitle footnotes`).

Let me know what you think!

jkrumbiegel:
Can you see if pinning makie to previous patch fixes the reftests?

stelmo:
Okay, I just checked, the tests pass with Makie v0.24.9. It seems the Makie update slightly changed some of the pixels in those two plots. For what it's worth, here are the diffs when running at Makie v.0.24.10. Should I update these reference pngs here?

`contour analysis`
image

`title subtitle footnotes`
image

jkrumbiegel:
Is the text more even in the new version or less? I wonder why it changed, maybe the CairoMakie glyph bundling PR

stelmo:
It is super hard to see unfortunately, here is a vector graphic zoom in of the "l" that changed in "false" in `contour analysis`:
image
And here is a similar zoom of the footnote difference:
image
It looks fine to me, but I am not really sure what counts as good enough... The pictures are for the new versions

jkrumbiegel:
I'll check it later when I have access to my computer. I just don't want to update refimages if this is a Makie regression. Maybe there's a good reason why we got a change here though.

stelmo:
For sure, but in any case it is probably better practice if the ref image change is not part of this PR, to keep the history nice and clean

jkrumbiegel:
Yes I would merge this one later

PR #766, opened 2026-06-02
fix(scales): Match palette and categories against presorted/nonnumeric data

jkrumbiegel:
When you wrap a column with `presorted` (or `nonnumeric`) and then want to assign colors via a palette, you have to specify the palette keys as plain values like `"Cat" => :green`. But the data values are `Presorted("Cat")`, so the palette lookup never matched and you'd get an error like

```
ERROR: ArgumentError: Key Fish not found and no default values are present
```

The same applied to `categories`, where listing the plain values failed against the wrapped data. This was the issue discussed in #720; rather than overloading `isequal` on the wrapper types (which is a fundamental operation better left alone), this adds a recursive `unwrap` and compares both the given and the existing value unwrapped before checking equality.

So this now works as you'd expect:

```julia
df = (; Pet = ["Fish", "Dog", "Cat", "Bird"], Ranking = 1:4)
spec = data(df) *
    mapping(:Pet => presorted, :Ranking, color = :Pet => presorted) *
    visual(BarPlot)

palette = ["Cat" => :green, "Dog" => :red, "Fish" => :blue, "Bird" => :yellow]
draw(spec, scales(Color = (; palette)))
```

bar plot with presorted pets colored via a plain-string palette

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Palette pairs and `categories` passed via `scales` now match `presorted`/`nonnumeric`-wrapped data without needing to re-wrap the keys [#766](https://github.com/MakieOrg/AlgebraOfGraphics.jl/pull/766).
+
 ## v0.12.11 - 2026-06-02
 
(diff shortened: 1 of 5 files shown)


TheLostLambda:
Hi @jkrumbiegel !

Sorry I didn't end up having the time for this — thank you for sorting it out!

jkrumbiegel:
You're welcome, and thanks for getting the ball rolling

PR #779, opened 2026-07-11
Support temporal columns on continuous color and size scales

jlbosse:
Previously, using `mapping(color = :timestamp)` where `timestamp` corresponds to a column containing `DateTime` resulted in comparing floats with DateTimes and errored in Makie's numbers_to_colors.

This PR fixes that by stripping temporal extrema via the existing datetime2float conversion at the same point where the Unitful extension strips units, and reuse DateTicksWrapper so colorbars and markersize/linewidth legends show date-formatted labels.

That enables using temporal columns on continuous color and size scales. 

**AI disclaimer**   
This PR was largely written by Claude, but I did check all of it and didn't see anything obviously wrong (as someone who is not super familiar with this code base, but familiar with good julia code).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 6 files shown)


jkrumbiegel:
Thanks for the PR! I've generalized the approach a bit in #781 as I needed some related fixes to datetime ticks etc.

